PR #2, opened 2015-02-20
limit dataframe deep copies in utils

erikvdp:


CamDavidsonPilon:
@erikvdp, thanks for these changes :shaved_ice:!

PR #4, opened 2015-02-25
fix import code in README

erikvdp:


CamDavidsonPilon:
Test failures are unrelated, and clearly something I need to address. Thanks for this change, @erikvdp!

PR #5, opened 2015-02-26
moved conditional_probability_alive_matrix function + possible bug fix

erikvdp:
Hi, I've moved the conditional_probability_alive_matrix function to the model. While doing so, I noticed that the matrix doesn't calculate all values, so I've added a row and column. It should be like that no?

### DIFF
--- a/lifetimes/plotting.py
+++ b/lifetimes/plotting.py
@@ -95,6 +95,7 @@ def plot_frequency_recency_matrix(model, max_x=None, max_t=None, **kwargs):
 
     return ax
 
+
 def plot_probability_alive_matrix(model, max_x=None, max_t=None, **kwargs):
     """
     Plot a figure of the probability a customer is alive based on their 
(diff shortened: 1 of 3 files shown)


lifetimes/plotting.py
```diff
@@ -73,9 +73,9 @@ def plot_frequency_recency_matrix(model, max_x=None, max_t=None, **kwargs):
         max_t = int(model.data['T'].max())
 
     t = 1  # one unit of time
-    Z = np.zeros((max_t, max_x))
```
CamDavidsonPilon: This function can now use the `conditional_probability_alive_matrix` you defined above. Want to use that here?

CamDavidsonPilon: oops, nvm. Please ignore this.

lifetimes/plotting.py
```diff
@@ -104,26 +105,17 @@ def plot_probability_alive_matrix(model, max_x=None, max_t=None, **kwargs):
...
-        max_x = int(model.data['frequency'].max())
-
-    if max_t is None:
-        max_t = int(model.data['T'].max())
-
-    Z = np.zeros((max_t, max_x))
-    for i, t_x in enumerate(np.arange(max_t)):
-        for j, x in enumerate(np.arange(max_x)):
-            Z[i, j] = model.conditional_probability_alive(x, t_x, max_t)
+    z = model.conditional_probability_alive_matrix(max_x, max_t)
```
CamDavidsonPilon: :+1:

CamDavidsonPilon:
You're correct about the `ranges`, nice catch. Circle keeps failing, but it's unrelated to these tests. Actually, can you add a test for the `conditional_probability_alive_matrix` too please?

erikvdp:
Hi again,
I wanted to check if this test is good enough for you.
I was planning to compare the entries of the <code>conditional_probability_alive_matrix</code> against the corresponding <code>conditional_probability_alive</code> functions, but they seem to be slightly off. Do you have any idea why that is?

CamDavidsonPilon:
My best guess is that it's an off by one error - can you reproduce the inconsistency?

erikvdp:
Hi, The small error in the <code> test_conditional_probability_alive_matrix </code> method was due to the T_max not properly rounded. It should be ok now.

CamDavidsonPilon:
Great, thanks for this contribution! Merging now.

PR #7, opened 2015-03-06
Added duration_cal column

erikvdp:
I've added duration_cal column to the dataframe that is returned by calibration_and_holdout_data function.

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -49,6 +49,7 @@ def calibration_and_holdout_data(transactions, customer_id_col, datetime_col, ca
     calibration_summary_data = summary_data_from_transaction_data(calibration_transactions, customer_id_col, datetime_col,
                                                                   datetime_format, observation_period_end=calibration_period_end, freq=freq)
     calibration_summary_data.columns = [c + '_cal' for c in calibration_summary_data.columns]
+    calibration_summary_data['duration_cal'] = to_period(calibration_period_end) - to_period(min(calibration_transactions[datetime_col]))
 
     # create holdout dataset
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
To confirm: this represents the duration between the observation period end and the earliest purchase - and it's the same for everyone? What are some use cases of this?

erikvdp:
This represents the actual duration of the calibration period(so time between the first purchase in the dataset up until the last purchase in the calibration period) . Which is in fact the same for everyone. I can imagine that it'll be useful in later stages when determining the length of calibration period for getting a good estimate on the observation period.

CamDavidsonPilon:
I don't think I like this addition. It is a redundant column if it's the same for everyone, and can be calculated outside of the function quite easily. It's also slightly misleading, as it looks like the individuals life so far (but it's really not). What are your thoughts? 

When I was looking at this PR, even though it was only 1 line addition, I was super confused about what the variables and columns represented. For example, at first glance, I thought "what is `T` vs `duration`?" There's probably a better API for this. 

Also, I appreciate your PRs and comments. Would you like to be a core contributor to lifetimes?

erikvdp:
> I don't think I like this addition. It is a redundant column if it's the same for everyone, and can be calculated outside of the function quite easily

True, however we have the same for the <code> duration_holdout </code> no? It is not possible for that column to have different values.

However it's indeed true that this can be calculated outside the function easily, it's totally fine for me to discard this pull request.

>Also, I appreciate your PRs and comments. Would you like to be a core contributor to lifetimes?
Thank you! I'm learning about the model on the go, but I'd be honored to be mentioned as contributor

PR #10, opened 2015-03-18
fixing some issues in plot_history_alive

CamDavidsonPilon:
cc @erikvdp for review please. 

I was playing around with the new plotting (with data from `load_transaction_data`) and encountered these bugs. Just to confirm, can you share your pandas version too? Mine is 0.15.1.

### DIFF
--- a/lifetimes/plotting.py
+++ b/lifetimes/plotting.py
@@ -192,7 +192,8 @@ def plot_history_alive(model, transactions, datetime_col, units, freq='D', **kwa
 
     # Get purchasing history of user
     customer_history = transactions[[datetime_col]].copy()
-    customer_history.index = customer_history[datetime_col]
+    customer_history.index = pd.DatetimeIndex(customer_history[datetime_col])
+
     # Add transactions column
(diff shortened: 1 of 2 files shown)


erikvdp:
Hi, my pandas version is 0.15.2.
The changes look good to me since you can't assume that the <code>datetime_col</code> is already in the right format, good catch. 

P.S. Can I ask you how your typical workflow looks when you are developing for lifetimes, or any other python library? I usually write my code/functions out in a python notebook before trying to incorporate them in the library (which is loaded as a project in my IDE).  However, I feel that this can easily lead to bugs like this. Do you have a better approach?

CamDavidsonPilon:
My approach is usually as follows:

 1. start in a text-editor with the library. Write a bit of code. If you get stuck with a some complicated pandas logic or some python syntax, go to the ipython shell and play around there to get it, then return to the text editor. 
 2. once done the first version, go to Ipython shell, import the new code, and play around with it. Is it as you expect?
 3. repeat 1 & 2 until happy. 
 4. write tests.

PR #22, opened 2015-08-28
CLV estimation

luke14free:
Hi Cam, here is my work on the Gamma/Gamma model. Let me know what you think


lifetimes/utils.py
```diff
@@ -143,16 +158,19 @@ def calculate_alive_path(model, transactions, datetime_col, t, freq='D'):
...
+        return function(params, *func_args)
 
     for i in range(iterative_fitting + 1):
         fit_method = methods[i % len(methods)]
-        params_init = np.random.exponential(0.5, size=4) if initial_params is None else initial_params
-        output = minimize(minimizing_function, method=fit_method, tol=1e-6,
-                          x0=params_init, args=(frequency, recency, T, penalizer_coef), options={'disp': disp})
+        params_init = np.random.exponential(0.5, size=params_size) if initial_params is None else initial_params
+        output = minimize(_func_caller, method=fit_method, tol=1e-6,
+                          x0=params_init, args=(minimizing_function_args, minimizing_function), options={'disp': disp})
```
CamDavidsonPilon: oh neat :+1:

lifetimes/estimation.py
```diff
@@ -8,6 +8,7 @@
 
 from scipy import special
 from scipy import misc
+from scipy.optimize import minimize
```
CamDavidsonPilon: needed?

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+            df['expected_revenues_period_'+str(i)] = df.apply(
+                lambda r: (m*transaction_prediction_model.predict(i, r['frequency'], r['recency'], r['T'])/(1+d)**(i/30)),
+                axis=1
+            )
+            self.discounted_monthly_cash_flows.append(df['expected_revenues_period_'+str(i)].sum())
+
+        return sum(self.discounted_monthly_cash_flows)
+
+
+    def fit(self, frequency, monetary_value, iterative_fitting=20, initial_params=None, verbose=False):
```
CamDavidsonPilon: 20 seems awfully high, can we lower this to 10?

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+        df['frequency'] = frequency
+        df['recency'] = recency
+        df['T'] = T
+        df['monetary_value'] = monetary_value
+
+        d = discount_rate
+        m = self.conditional_expected_average_profit()
+        self.discounted_monthly_cash_flows = []
+
+        for i in range(30, (time*30)+1, 30):
```
CamDavidsonPilon: what is 30 here? Maybe better as a parameter

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+            df['expected_revenues_period_'+str(i)] = df.apply(
+                lambda r: (m*transaction_prediction_model.predict(i, r['frequency'], r['recency'], r['T'])/(1+d)**(i/30)),
```
CamDavidsonPilon: `i/30` worries me in Python2. How about `from __future__ import division` at the top?

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+
+        Returns:
+            the conditional expectation of the average profit per transaction
+        """
+        m = self.data['monetary_value'] if m is None else m
+        x = self.data['frequency'] if x is None else x
+        p, q, v = self.params_.values()
+        return np.mean((((q - 1) / (p * x + q - 1)) * (v * p / (q - 1))) + (p * x / (p * x + q - 1)) * m)
+
+    def customer_lifetime_value(self, transaction_prediction_model, frequency, recency, T, monetary_value, time=12, discount_rate=1):
```
CamDavidsonPilon: Correct me if I am wrong here, but this is for a single customer?

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+                frequencies used for fitting the model.
+            m: a vector containing the customers' monetary values. Defaults to the whole set of
+                monetary values used for fitting the model.
+
+        Returns:
+            the conditional expectation of the average profit per transaction
+        """
+        m = self.data['monetary_value'] if m is None else m
+        x = self.data['frequency'] if x is None else x
+        p, q, v = self.params_.values()
```
CamDavidsonPilon: `p, q, v = self._unload_params('p', q', 'v')` to guarantee order

lifetimes/estimation.py
```diff
@@ -37,6 +38,117 @@ def _print_params(self):
...
+
+        penalizer_term = penalizer_coef * log(params).sum()
+        negative_log_likelihood = np.sum(negative_log_likelihood_values) + penalizer_term
+
+        if np.any(np.isnan(negative_log_likelihood_values)) or np.isnan(negative_log_likelihood):
+            return np.inf
+
+        return negative_log_likelihood
+
+    def conditional_expected_average_profit(self, x=None, m=None):
```
CamDavidsonPilon: For example, if a customer has frequency 10, and purchase value $100, this compute `E[P | x=10, m=100]` where P is future profit. Is _future_ profit correct?

README.md
```diff
@@ -209,6 +210,73 @@ our trained model. For example:
...
+
+For this whole time we didn't take into account the economic value of each transaction and we focused mainly on
+transactions' occurences. To estimate this we can use the Gamma/Gamma model. But first we need to create summary data
+from transactional data also containing economic values for each transaction (i.e. profits or revenues).
+
+    from lifetimes.datasets import load_transaction_data_with_monetary_value
+
+    summary_with_money_value = load_transaction_data_with_monetary_value()
+    summary_with_money_value.head()
+    returning_customers_summary = summary_with_money_value[summary_with_money_value['frequency']>0]
```
CamDavidsonPilon: I guess it's in the literature, but why does `['frequency']>0`?

README.md
```diff
@@ -209,6 +210,73 @@ our trained model. For example:
...
+ 
+    print ggf.customer_lifetime_value(
+        bgf, #the model to use to predict the number of future transactions
+        returning_customers_summary['frequency'],
+        returning_customers_summary['recency'],
+        returning_customers_summary['T'], 
+        returning_customers_summary['monetary_value'], 
+        time=12, 
+        discount_rate=0.7
+    ) #10689.30
```
CamDavidsonPilon: :+1: nice docs!

CamDavidsonPilon:
Thank you for this PR! Things are looking good. Can I ask you to add some tests, especially around `summary_data_from_transaction_data`, and if possible to test the output of `GammaGammaFitter` here to an existing implementation (or just as good, values reported in literature). 

Let me know if you have any questions!

lifetimes/datasets/__init__.py
```diff
@@ -43,3 +45,23 @@ def load_transaction_data(**kwargs):
...
+    )
+
+    transactions['date'] = pd.to_datetime(transactions['date'], format='%Y%m%d')
+
+    return utils.summary_data_from_transaction_data(
+        transactions,
+        'customer_id',
+        'date',
+        monetary_value_col='dollar_value',
+        observation_period_end=min(transactions['date'])
```
CamDavidsonPilon: I think this should be `max(transactions['date']`

lifetimes/datasets/__init__.py
```diff
@@ -43,3 +45,23 @@ def load_transaction_data(**kwargs):
...
+def load_transaction_data_with_monetary_value(**kwargs):
+    transactions = pd.read_csv(
+        resource_filename(
+            'lifetimes', 'datasets/CDNOW_master.txt'
+        ),
+        sep='[\t|\s]+',
+        engine='python', **kwargs
+    )
+
+    transactions['date'] = pd.to_datetime(transactions['date'], format='%Y%m%d')
```
CamDavidsonPilon: This can be done in the call to `summary_data_from_transaction_data` by setting `datetime_format='%Y%m%d'`

CamDavidsonPilon:
Please add `GammaGammaFitter` here: https://github.com/CamDavidsonPilon/lifetimes/pull/22/files#diff-3810f074613cf8b640b39ed950d552f5R16

CamDavidsonPilon:
Pulling this locally, it is quite unstable. I think a major issue is in the calculations of `p_0` and `p_1`. Consider this pseudo-code / pseudo-math

```
 negative_log_likelihood_values = (-np.log(p_0 * p_1))
                                                    = -np.log(p_0)  - np.log(p_1)
```
`-np.log(p_0)` is the logarithm of a ratio of gammas, so it can be expanded to sums and subtractions of the logarithm of gammas. There exists a numerically stable `gammaln` in scipy (I use it in the other fitters. 

Similarly, `-np.log(p_1)` is currently the product of exponentials. This is unstable for numbers far away from 0, so I would suggest expanding it into a summation.

lifetimes/estimation.py
luke14free: This is a bit tricky. In finance you have something which is called "discounted cash flow", which is a way to show that money tomorrow doesn't equal money today. Basically you discount (i.e. divide) future earnings at a discount rate (i.e. (1+d)^n, where d = monthly discount rate, n = month in which you earn that money). The most used interest rate is the monthly one (hence I divide the days used to figure out the number of future expected purchases by 30 -> average number of days in a month). In case we wanted to change it to the annual rate then the relationship is: monthly rate = (1 + annual rate)^(1/12) - 1. Let me know your take on this

lifetimes/estimation.py
luke14free: not needed since i is range(30, (time*30)+1, 30) so basically [30, 60, 90...] third parameter in range defines the step, so i can always be integer divided by 30

lifetimes/estimation.py
luke14free: m and x can be vectors containing 1..n values. so this is for any amount of customers. should I make it more clear?

lifetimes/estimation.py
luke14free: The Gamma/Gamma submodel estimates the average profit per transaction of a customer and assumes it is constant over time. So I believe that future = present and your affirmation should be correct.

README.md
luke14free: Literature (Fader et al. 2005) uses x > 0, but the main reason I do it now is that while computing p_0 I need to compute `np.log(special.gamma(x))` which would be +inf if freq was to be 0. Otherwise I could be using laplacian smoothing (e.g. adding a +1 to all the frequencies, but I am not sure if this is a sound mathematical choice, what do you think?).

lifetimes/datasets/__init__.py
luke14free: (Y)

lifetimes/estimation.py
luke14free: agreed

lifetimes/utils.py
luke14free: ;-)

lifetimes/estimation.py
```diff
@@ -37,6 +37,117 @@ def _print_params(self):
...
+    def _negative_log_likelihood(params, frequency, avg_monetary_value, penalizer_coef):
+        if any(i < 0 for i in params):
+            return np.inf
+
+        p, q, v = params
+
+        x = frequency.astype('int')
+        m = avg_monetary_value.astype('int')
+
+        p_0 = np.log(special.gamma(p*x+q)) - np.log(special.gamma(p)) - np.log(special.gamma(x)) - np.log(special.gamma(q))
```
CamDavidsonPilon: I think there is an error here. Correct me if I am wrong. The original code you had was:
```
special.gamma((p * x) + q) / (special.gamma(p * x) * special.gamma(q))
```

taking the log of this:

```
np.log(special.gamma((p * x) + q)) - np.log(special.gamma(p * x)) - np.log(special.gamma(q))
```

but you have separated the third term into `np.log(special.gamma(p)) - np.log(special.gamma(x))`.

CamDavidsonPilon: Also, using `gammaln(*)` in scipy: http://docs.scipy.org/doc/scipy/reference/generated/scipy.special.gammaln.html is more stable than `np.log(special.gamma(*))`

lifetimes/estimation.py
```diff
@@ -37,6 +37,117 @@ def _print_params(self):
...
+        p_1 = q*np.log(v) + (p*x-1)*np.log(m) + p*x*np.log(x) - (p*x+q)*np.log(v + m*x)
```
CamDavidsonPilon: Ditto here: I believe there is a mistake. I get: 

```
np.log((v ** q) + (m ** ((p * x) - 1)) * (x ** (p * x))) - ((p * x) + q) * np.log(v + (m * x))

```

Can you double check apply logarithms here?

lifetimes/estimation.py
CamDavidsonPilon: I think making it a parameter in the function call would make sense, with default 30 is fine.

lifetimes/estimation.py
CamDavidsonPilon: okay cool

README.md
CamDavidsonPilon: no it makes sense given the model to set x>0. Averages of a single number are a bit of a redundant pathology in statistics.

CamDavidsonPilon:
Looking good, but I'd still like to see some tests added =)

lifetimes/estimation.py
luke14free: So basically p_1 is this:

screen shot 2015-09-08 at 10 24 00 am

if you check wolframalpha for the log equivalent of it http://www.wolframalpha.com/input/?i=log%28+%5B+%28v%5Eq%29+*+%28+m%5E%28p*x-1%29+*+x%5E%28p*x%29+%29+%5D+%2F+%5B+%28v+%2B+m*x%29%5E%28p*x-1%29+%5D+%29+ 

you get

screen shot 2015-09-08 at 10 33 59 am

That is equivalent to
```
p_1 = q\*np.log(v) + (p\*x-1)\*np.log(m) + p\*x\*np.log(x) - (p\*x+q)\*np.log(v + m\*x)
```

lifetimes/estimation.py
luke14free: what a stupid mistake. my bad. gammaln is a super idea.

luke14free:
@CamDavidsonPilon how did you create the cdnow_customers.csv file? I wanted to add monetary values to it so to create the appropriate tests but I don't understand the logic you used to create it so to reproduce Hardie's results.

lifetimes/estimation.py
CamDavidsonPilon: :+1: awesome, thanks for checking that!

CamDavidsonPilon:
So in the gamma-gamma paper by Hardie, he mentions an excel sheet, located here: http://brucehardie.com/notes/025/

The data is contained in there, you can pull that data out, save it, and add it as a file in this PR (located in `/lifetimes/datasets`)

luke14free:
While creating the tests I noticed something weird:

```
summary_with_money_value = load_transaction_data_with_monetary_value()
returning_customers_summary = summary_with_money_value#[summary_with_money_value['frequency']>0]

print len(returning_customers_summary) #946, exactly as in Hardie's

ggf = GammaGammaFitter(penalizer_coef = 0)
ggf.fit(returning_customers_summary['frequency'], 
        returning_customers_summary['monetary_value'], iterative_fitting=2)

pars = dict(ggf.params_)
p = pars['p']
q = pars['q']
v = pars['v']

print "My target f. value", ggf._negative_log_likelihood((p,q,v), returning_customers_summary['frequency'], 
        returning_customers_summary['monetary_value'], 0) #My target f. value 4058.09170696

print "Hardie's target f. value", ggf._negative_log_likelihood((6.25,3.74,15.33), returning_customers_summary['frequency'], 
        returning_customers_summary['monetary_value'], 0) #Hardie's target f. value 4060.15263452
```

I also double checked my full MLE function with WA (http://www.wolframalpha.com/input/?i=log%28+%5B+gamma%28p*x+%2B+q%29+%2F+%28gamma%28p*x%29+*+gamma%28q%29%29+%5D+%5B+v%5Eq+*++m%5E%28p*x-1%29+*+x%5E%28p*x%29+%5D+%2F+%5B+%28v+%2B+m*x%29%5E%28p*x%2Bq%29+%5D+%29+) and rewrote my function as:

```
negative_log_likelihood_values = (-1+p*x)*log(m)+q*log(v)+p*x*log(x)+(-q-p*x)*log(v+m*x)\
                                         -special.gammaln(q)-special.gammaln(p*x)+special.gammaln(q+p*x) 
```

It seems like we found a better minima than the one found in literature. Maybe it's because they used excel for fitting the curve?

CamDavidsonPilon:
> It seems like we found a better minima than the one found in literature.

That possible. Alternatively, can you confirm the summary statistics that Hardie presents in the paper are the same as in your dataset? Maybe the data is different.

screen shot 2015-09-08 at 09 45 50

luke14free:
It seems the same data. Furthermore I also get the 11% correlation he describes in Fader, Hardie 2005.

count    946.000000
mean      35.078552
std       30.283514
min        2.990000
25%       15.760000
50%       27.500000
75%       41.785000
max      299.630000

CamDavidsonPilon:
Do they outputted coefficients vary from the paper?

luke14free:
I double checked Hardie's xls and it was a floating point error. Casting m and x to float 64 solves it and now we arrive to his same coefficients. Sorry for the long time it took to fix this PR, it was one of my first MLE optimization contribution, I am sure that the next will go smoother :+1:

luke14free:
(tests on their way)

CamDavidsonPilon:
Great work debugging!

CamDavidsonPilon:
:+1: nice job, I just wanna take it for a spin before I merge =)

lifetimes/datasets/__init__.py
```diff
@@ -43,3 +45,11 @@ def load_transaction_data(**kwargs):
 
     """
     return load_dataset('example_transactions.csv', **kwargs)
+
+
+def load_transaction_data_with_monetary_value(**kwargs):
+    df = load_dataset('cdnow_customers_transactions.csv', **kwargs)
+    df.columns = ['customer_id', 'frequency', 'recency', 'T', 'monetary_value']
+    df.index = df['customer_id']
+    del df['customer_id']
```
CamDavidsonPilon: lines 53-54 can be accomplished more cleanly with
```
df = df.set_index('customer_id')
```

lifetimes/datasets/__init__.py
```diff
@@ -43,3 +45,11 @@ def load_transaction_data(**kwargs):
 
     """
     return load_dataset('example_transactions.csv', **kwargs)
+
+
+def load_transaction_data_with_monetary_value(**kwargs):
```
CamDavidsonPilon: This name is misleading as it suggests transactional data. Something like: `load_summary_data_with_monetary_value` sounds better

README.md
```diff
@@ -209,6 +210,73 @@ our trained model. For example:
 
 ![history](http://i.imgur.com/y45tum4.png)
 
+### Estimating Customers' Lifetime Value
+
+For this whole time we didn't take into account the economic value of each transaction and we focused mainly on
+transactions' occurences. To estimate this we can use the Gamma/Gamma model. But first we need to create summary data
```
CamDavidsonPilon: `occurences` -> `occurrences`

CamDavidsonPilon: You use `Gamma-Gamma` below, mind changing it here?

README.md
```diff
@@ -209,6 +210,73 @@ our trained model. For example:
...
+    ggf.fit(returning_customers_summary['frequency'], 
+            returning_customers_summary['monetary_value'])
+    print ggf    
+    """
+    <lifetimes.GammaGammaFitter: fitted with 9591 subjects, p: 163.91, q: 7.36, v: 0.06>
+    """
+    
+We can now produce the average CLV figure for our dataset as follows:
+    
+    print "Epected conditional average profit: %s, Average profit: %s" % (
```
CamDavidsonPilon: `Expected`, and same below

lifetimes/estimation.py
```diff
@@ -37,6 +37,109 @@ def _print_params(self):
...
+        """
+        df = DataFrame()
+        df['frequency'] = frequency
+        df['recency'] = recency
+        df['T'] = T
+        df['monetary_value'] = monetary_value
+
+        d = discount_rate
+        m = self.conditional_expected_average_profit()
+        self.discounted_monthly_cash_flows = []
```
CamDavidsonPilon: do we really want this attached to the object? Do we use it later?

CamDavidsonPilon:
There's a errant file `lifetimes/showDeclFileRes.do?declId=8555` in the PR too

lifetimes/estimation.py
```diff
@@ -37,6 +37,109 @@ def _print_params(self):
...
+
+        Returns:
+            the conditional expectation of the average profit per transaction.
+            Also creates a discounted_monthly_cash_flows attribute
+        """
+        df = DataFrame()
+        df['frequency'] = frequency
+        df['recency'] = recency
+        df['T'] = T
+        df['monetary_value'] = monetary_value
```
CamDavidsonPilon: Is this used in the calculation below?

lifetimes/estimation.py
```diff
@@ -37,6 +37,109 @@ def _print_params(self):
...
+            m: a vector containing the customers' monetary values. Defaults to the whole set of
+                monetary values used for fitting the model.
+
+        Returns:
+            the conditional expectation of the average profit per transaction
+        """
+        m = self.data['monetary_value'] if m is None else m
+        x = self.data['frequency'] if x is None else x
+        p, q, v = self._unload_params('p', 'q', 'v')
+        return np.mean((((q - 1) / (p * x + q - 1)) * (v * p / (q - 1))) + (p * x / (p * x + q - 1)) * m)
```
CamDavidsonPilon: Where is this formula from?

lifetimes/estimation.py
```diff
@@ -37,6 +37,109 @@ def _print_params(self):
...
+
+        negative_log_likelihood_values = special.gammaln(p*x+q)-special.gammaln(p*x)-special.gammaln(q)\
+                                         +q*np.log(v)+(p*x-1)*np.log(m)+(p*x)*np.log(x)-(p*x+q)*np.log(x*m+v)
+
+        penalizer_term = penalizer_coef * log(params).sum()
+        negative_log_likelihood = -np.sum(negative_log_likelihood_values) + penalizer_term
+
+        return negative_log_likelihood
+
+    def conditional_expected_average_profit(self, x=None, m=None):
```
CamDavidsonPilon: Can you remove the kwargs to `frequency` and `monetary_value` to be consistent?

lifetimes/estimation.py
CamDavidsonPilon: Also, does it make more sense for this to be applied row-wise. ATM, if I enter: 
```
ggf.conditional_expected_average_profit(returning_customers_summary['frequency'].ix[:10], returning_customers_summary['monetary_value'].ix[:10])
```

I get a single value back. I feel like I should instead get the `conditional_expected_average_profit` per customer, essentially returning an array.  What do you think?

luke14free: So this is the formula to compute the expected money value for any transaction, that is later used in the pareto-independent models to compute the deviation of people transactions from the expected one. Basically it takes many transactions and it averages them in a weighted way to produce m-tilda. Also Hardie arrives to a single value.

Formula comes from: http://www.essec.edu/faculty/showDeclFileRes.do?declId=8555&key=Publication-Content (m-hat 3.12 page 7).

I am fixing the kwargs as suggested

lifetimes/estimation.py
luke14free: Yes I use it to do compute the cashflows per period by applying the expected conditional number of future purchases by row which is particularly handy. Do you think it's better to del the df explicitly and call the GC after I am done with it to free up memory?

lifetimes/estimation.py
luke14free: In general it's nice to have the DCF as pieces because maybe we want to plot it or show it in some way. But this is just my personal preference.

lifetimes/estimation.py
CamDavidsonPilon: What I mean is, I don't see where `monetary_value` is used again in this function.

lifetimes/estimation.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 from numpy import log, exp, logaddexp, asarray, any as npany, c_ as vconcat,\
                   isinf, isnan, ones_like
-from pandas import DataFrame
+from pandas import DataFrame, ols
```
CamDavidsonPilon: `ols` is not used, is it?

luke14free: yes, sorry it comes from a branch where i am trying to implement the pareto independent model in which i have to use it.

CamDavidsonPilon:
Cool, after https://github.com/CamDavidsonPilon/lifetimes/pull/22#discussion_r39353809 is looked at, good to merge!

luke14free:
:-) sorry for the delay, we should be good to go!

CamDavidsonPilon:
Apologies for taking so long - I thought I merged this! Is this still good to go?

luke14free:
Yes :-)

— 
Luca Giacomel
CEO, PolicyBrain s.r.l. 
Tel. +390236571091

PR #37, opened 2016-01-17
Made some plotting calls compatible with matplotlib >=1.5

rhydomako:
Some of the plotting functions were broken for matplotlib >=1.5, because of some changes to matplotlibs internal API. This PR makes some changes to try to use the public-facing matplotlib API where applicable, or make version-dependent calls if necessary.

I have also added a test for plot_probability_alive_matrix.

### DIFF
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -37,7 +37,7 @@ class TestPlotting():
         plt.show()
 
     def test_plot_probability_alive_matrix(self):
-        print plt.matplotlib.__version__
+        from matplotlib import pyplot as plt
 
         plt.figure()
         plotting.plot_probability_alive_matrix(bgf)
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -36,6 +36,17 @@ class TestPlotting():
 
         plt.show()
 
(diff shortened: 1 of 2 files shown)


rhydomako:
I should note that I tested this with matplotlib 1.4.3, 1.5.0, and 1.5.1.

tests/test_plotting.py
```diff
@@ -36,6 +36,17 @@ def test_plot_frequency_recency_matrix(self):
 
         plt.show()
 
+    def test_plot_probability_alive_matrix(self):
+        print plt.matplotlib.__version__
```
CamDavidsonPilon: We are compatible with Python3, so this syntax won't work. In other files, we use 
```
from __future__ import print_function
```
in the head of the file, so we can use `print()`.

CamDavidsonPilon:
Terrific, thanks for fixing this. One comment, and otherwise lgtm

tests/test_plotting.py
rhydomako: Whoops, I didn't want that print to survive at all. I've changed it to the import that I was actually going for -- so that it can be run as a standalone test, and to be consistent with the other tests.

PR #38, opened 2016-01-19
iteritems() has been removed in python 3.x

rhydomako:
Orderdict.items() should be compatible (at least in this case) with both python 2 and 3.


CamDavidsonPilon:
:+1: nice catch, @rhydomako!

PR #40, opened 2016-01-24
Fix issues with summary_data_from_transaction_data() when including monetary_values column

rhydomako:
I've found two issues with the function summary_data_from_transaction_data() when including an optional monetary_values column:

1. transaction in the same period aren't reduced (summed) together,
2. all transactions are included in the average purchase amount, when generally these model are only concerned with repeated transactions/purchases.

The PR refactors summary_data_from_transaction_data() to address both of these issues (I refer to http://brucehardie.com/notes/022/RFM_summary_in_Excel.pdf and http://brucehardie.com/notes/025/gamma-gamma_calibration.xlsx for how I'm expecting monetary values to be processed).

I've also introduced a function find_repeated_transactions(), which explicitly groups transactions by period and labels repeated transactions. I've separated this functionality from summary_data_from_transaction_data() because I expect that it will be useful for other things like comparing model/actual transaction volume as a function of time.

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -7,7 +7,7 @@ from scipy.optimize import minimize
 pd.options.mode.chained_assignment = None
 
 __all__ = ['calibration_and_holdout_data',
-           'find_repeated_transactions',
+           'find_first_transactions',
            'summary_data_from_transaction_data',
            'calculate_alive_path']
 
@@ -71,8 +71,12 @@ def reduce_events_to_period(transactions, *aggregation_columns):
     return transactions.groupby(aggregation_columns, sort=False).agg(lambda r: 1)
 
 
-def find_repeated_transactions(transactions, customer_id_col, datetime_col, monetary_value_col=None, datetime_format=None,
-                               observation_period_end=datetime.today(), freq='D'):
+def reduce_events_to_period(transactions, *aggregation_columns):
+    return transactions.groupby(aggregation_columns, sort=False).agg(lambda r: 1)
+
+
+def find_first_transactions(transactions, customer_id_col, datetime_col, monetary_value_col=None, datetime_format=None,
+                            observation_period_end=datetime.today(), freq='D'):
     """
     This takes a Dataframe of transaction data of the form:
         customer_id, datetime [, monetary_value]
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
Ah sorry I haven't looked at this yet. I promise I will this week :)

CamDavidsonPilon:
@rhydomako does this PR do more than what https://github.com/CamDavidsonPilon/lifetimes/pull/42 does? cc @jackyckma

rhydomako:
I think #42 fixes my issue 1, but not issue 2. That is, I think only the repeated purchases should be included in the mean purchase amount. I've added a dedicated function to identify repeated transactions to address this issue.

lifetimes/utils.py
```diff
@@ -101,31 +98,93 @@ def summary_data_from_transaction_data(transactions, customer_id_col, datetime_c
...
-    period_transactions = reduce_events_to_period(transactions, *select_columns).reset_index(level=select_columns[1:])
+    # join the first_transaction column to the transaction log
+    period_transactions = pd.merge(
+        period_transactions,
+        first_transactions[[customer_id_col, datetime_col, 'first']],
+        how="left",
+        on=[customer_id_col, datetime_col]
+    )
+    # boolean invert the 'first' column to give a column representing repeated transactions
+    period_transactions['repeated'] = ~(period_transactions['first'].fillna(False))
```
CamDavidsonPilon: I'm not crazy about how this is computed, that is, it doesn't feel very clear to me. What about something like:

```
def _set_repeated_in_df(df):
    df['repeat'] = True
    df['repeat'].iloc[0] = False
    return df

period_transactions.sort_values(datetime_col).groupby(customer_id_col).apply(_set_repeated_in_df)
```

CamDavidsonPilon: I believe this is logically equal

lifetimes/utils.py
```diff
@@ -101,31 +98,93 @@ def summary_data_from_transaction_data(transactions, customer_id_col, datetime_c
...
+    # group together all transactions by customer_id and find each customers first and last transaction
+    customers = repeated_transactions.groupby(customer_id_col)[datetime_col].agg(['min', 'max'])
+    # count up all the repeated transactions for each customer
+    # by using a pivot_table, customers with zero repeated transactions are naturally accounted for
+    customers['frequency'] = pd.pivot_table(
+        repeated_transactions,
+        index=customer_id_col,
+        values='repeated',
+        aggfunc=np.sum
+    )
```
CamDavidsonPilon: what does this do? I'm just surprised that ` pd.pivot_table` returns a 1d array

tests/test_utils.py
```diff
@@ -71,6 +71,74 @@ def large_transaction_level_data_with_monetary_value():
     return pd.DataFrame(d, columns=['id', 'date', 'monetary_value'])
 
 
+def test_find_repeated_transactions_returns_correct_results(large_transaction_level_data):
```
CamDavidsonPilon: :+1: thanks for the tests!

lifetimes/utils.py
rhydomako: Interestingly, when I try your code, it ends up being like 2x slower. I think it is likely due to the apply being applied individually to all rows. I'll try to come up with some middle ground...

rhydomako: I've refactored this part to use an assignment by index. This should be both performant and easier to follow.

lifetimes/utils.py
rhydomako: I've changed the frequency calculation back to `customers['frequency'] = customers['count'] - 1` because it is certainly easier to understand. Similarly, I refactored out the pivot_table calculation for `monetary_value` because that was also a bit hard to follow.

rhydomako:
I spent a bit of time trying to use more vectorized functions and indexes in place of `apply` calls, which I think helped the overall readability. As well, at least on my machine, this seems to speed up `summary_data_from_transaction_data` by something like 50%.

CamDavidsonPilon:
Amazing, that function was very slow, so great job for speeding it up. :+1: merging now.

harrisong:
@rhydomako Thanks for this great PR. I was looking into why `summary_data_from_transaction_data` does not take first_purchases into the mean of the monetary value for the customers. Wouldn't that skew the CLV distribution? (Given that non-repeating customers should contribute to lower CLV)?

https://github.com/CamDavidsonPilon/lifetimes/pull/40/files#diff-48a7adefd56b3316f39206b07dadd6f5R181

rhydomako:
@harrisong Ok, this is a good question. The straightforward answer is that this is the way it is done in the original paper (see the XLSX in http://www.brucehardie.com/notes/025/ and http://brucehardie.com/notes/022/RFM_summary_in_Excel.pdf). Both the frequency and mean monetary-value per period are defined for repeat transactions only. The authors don't come out and say it, but my impression is that they've defined it this way to stay consistent across the board in terms of dropping first as well as singleton transactions. Dropping singleton transactions makes sense because without repeat transactions you can't really make any estimation of a transaction frequency, and I think dropping first transactions kinda followed in symmetry with that.

The corollary question is 'Is this a good idea when trying to estimate the mean monetary value of a user's transactions?' To be honest, I'm not sure. Like you say, we're dropping information that could be useful for estimating the distribution of transaction values. The Gamma-Gamma paper (http://www.brucehardie.com/notes/025/gamma_gamma.pdf) also just talks about user's mean transaction values and doesn't include any discussion about dropping transactions.

In general, I haven't seen as much validation of the Gamma-Gamma model, so I don't hold a strong opinion about its estimations. I kinda wonder if a simulation study might help give some insights into wether dropping these values throws a lot of bias into the estimates.

PR #42, opened 2016-01-27
fix inconsistent results from summary_data_from_transaction_data

jackyckma:
The summary result is different when monetary_value is included, which
caused by the groupby operation with select_columns which includes the
monetary_value column. Fixed by rewriting the groupby statement. The
calculation for monetary_value is updated as period_transaction is not
created before creating the summary.

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -109,11 +109,12 @@ def summary_data_from_transaction_data(transactions, customer_id_col, datetime_c
 
     transactions = transactions.ix[transactions[datetime_col] <= observation_period_end]
 
-    # reduce all events per customer during the period to a single event:
-    period_transactions = reduce_events_to_period(transactions, *select_columns).reset_index(level=select_columns[1:])
-
     # count all orders by customer.
-    customers = period_transactions[datetime_col].groupby(level=customer_id_col, sort=False).agg(['max', 'min', 'count'])
+    customers = transactions.groupby(customer_id_col).apply(
+        lambda x: pd.Series({'max': x[datetime_col].max(),
+                            'min': x[datetime_col].min(),
+                            'count': len(x[datetime_col].unique())
+                            }))
 
     # subtract 1 from count, as we ignore their first order.
     customers['frequency'] = customers['count'] - 1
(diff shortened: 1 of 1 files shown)


lifetimes/utils.py
```diff
@@ -109,11 +109,12 @@ def to_period(d):
...
 
-    # reduce all events per customer during the period to a single event:
-    period_transactions = reduce_events_to_period(transactions, *select_columns).reset_index(level=select_columns[1:])
-
     # count all orders by customer.
-    customers = period_transactions[datetime_col].groupby(level=customer_id_col, sort=False).agg(['max', 'min', 'count'])
+    customers = transactions.groupby(customer_id_col).apply(
+        lambda x: pd.Series({'max': x[datetime_col].max(),
+                            'min': x[datetime_col].min(),
+                            'count': len(x[datetime_col].unique())
```
CamDavidsonPilon: what are the logic differences between these lines?

lifetimes/utils.py
```diff
@@ -124,8 +125,8 @@ def to_period(d):
     summary_columns = ['frequency', 'recency', 'T']
 
     if monetary_value_col:
-        customers['monetary_value'] = period_transactions.groupby(level=customer_id_col)[
-            monetary_value_col].mean()
+        period_transaction = transactions.groupby([customer_id_col, datetime_col])[monetary_value_col].sum().reset_index()
+        customers['monetary_value'] = period_transaction.groupby(customer_id_col).mean()
```
CamDavidsonPilon: :+1: these changes make sense to me

lifetimes/utils.py
jackyckma: it's the same when *select_columns is just [customer_id_col, datetime_col]

but when *select_columns is [customer_id_col, datetime_col, monetary_value_col], the original statement will include monetary_value_col in the group such that 'same date different amount' transactions will not be grouped - which I believe that is not what it intends to be.

As I haven't constructed the period_transactions dataframe here, I just rewrite the groupby statement, and since simply `agg('count')` won't group the transactions in the same date as a single transaction, therefore I replace it with `lambda x: len(x[datetime_col].unique())`.

CamDavidsonPilon:
@jackyckma since #40 merged, can this PR be closed?

PR #45, opened 2016-01-29
Added the ModifiedBetaGammaFitter class and tests

rhydomako:
This PR would add an implementation of the Modified BG/NBD model. See http://research.sabanciuniv.edu/20/1/stvkaf01939.doc

This model is very similar to the BG/NBD model except that it has the customers sampling dropout their drop-out probabilities immediately after their first transaction instead of after a repeat transaction. This means that P(alive) shouldn't be stuck at 1 for one-time only buyers. Apart from that difference, most of the estimates should be very similar to the BG/NBD model.

Due to the similarity between the two models, for this implementation I've reused a lot of the BetaGammaFitter code with modified coefficients. This ends up feeling like there is a lot of repetition and it is definitely a question if it would make sense to combine the two classes, or just leave them separated.

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -12,7 +12,7 @@ from scipy import misc
 from lifetimes.utils import _fit, _scale_time, _check_inputs
 from lifetimes.generate_data import pareto_nbd_model, beta_geometric_nbd_model
 
-__all__ = ['BetaGeoFitter', 'ParetoNBDFitter', 'GammaGammaFitter']
+__all__ = ['BetaGeoFitter', 'ParetoNBDFitter', 'GammaGammaFitter', 'ModifiedBetaGeoFitter']
 
 
 class BaseFitter():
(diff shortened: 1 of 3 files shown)


CamDavidsonPilon:
Yes, the code is very similar. It probably makes sense to have `ModifiedBetaGeoFitter` a subclass of `BetaGeoFitter`, and only to reimplement the specific fitting/coef functions. (That is all the is different, right? As far as I can see, only a few methods are different)

rhydomako:
I've made `ModifiedBetaGeoFitter` inherit from `BetaGeoFitter`, and tried to make explicit the places where it differs from its parent class vs where it entirely replicates its behaviour. Oh, and yes, the interface is identical to `BetaGeoFitter`, with the differences being contained in the model coefficients.

tests/test_estimation.py
```diff
@@ -171,6 +171,109 @@ def test_scaling_inputs_gives_same_or_similar_results(self):
...
+        params = [1,1,1,1]
+        assert mbgf._negative_log_likelihood(params, np.array([x[0]]), np.array([t_x[0]]), np.array([t[0]]), 0) \
+             + mbgf._negative_log_likelihood(params, np.array([x[1]]), np.array([t_x[1]]), np.array([t[1]]), 0) \
+            == mbgf._negative_log_likelihood(params, x, t_x, t, 0)
+ 
+    def test_params_out_is_close_to_BTYDplus(self):
+        """ See https://github.com/mplatzer/BTYDplus """
+        mbfg = estimation.ModifiedBetaGeoFitter()
+        mbfg.fit(cdnow_customers['frequency'], cdnow_customers['recency'], cdnow_customers['T'], iterative_fitting=3)
+        expected = np.array([0.525, 6.183, 0.891, 1.614])
```
CamDavidsonPilon: :+1: nice!

lifetimes/estimation.py
```diff
@@ -476,3 +476,150 @@ def probability_of_n_purchases_up_to_time(self, t, n):
...
+            verbose: set to true to print out convergence diagnostics.
+
+
+        Returns:
+            self, with additional properties and methods like params_ and predict
+
+        """
+        super(self.__class__, self).fit(frequency, recency, T, iterative_fitting, initial_params, verbose) #although the partent method is called, this class's _negative_log_likelihood is referenced
+        self.generate_new_data = lambda size=1: modified_beta_geometric_nbd_model(T, *self._unload_params('r', 'alpha', 'a', 'b'), size=size) #this needs to be reassigned from the parent method
+        return self
```
CamDavidsonPilon: coolio, lgtm

CamDavidsonPilon:
Great work @rhydomako! I learned about the MBG/NBD thanks to you too


Merging, and I think I'll bump the version this week too :smile:

PR #47, opened 2016-02-04
fix clv calculation

jackyckma:
update CLV calculation:
- use individual’s monetary_value instead of
conditional_expected_average_profit for the whole training data.
- for each period, it should be the difference between
transaction_prediction_model.predict for successive periods instead of
the original one.

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -96,14 +96,17 @@ class GammaGammaFitter(BaseFitter):
         df['frequency'] = frequency
         df['recency'] = recency
         df['T'] = T
+        df['monetary_value'] = monetary_value
 
         d = discount_rate
-        m = self.conditional_expected_average_profit()
+
         discounted_monthly_cash_flows = []
 
         for i in range(30, (time*30)+1, 30):
             df['expected_revenues_period_'+str(i)] = df.apply(
-                lambda r: (m*transaction_prediction_model.predict(i, r['frequency'], r['recency'], r['T'])/(1+d)**(i/30)),
+                lambda r: (r['monetary_value']*
+                    (transaction_prediction_model.predict(i, r['frequency'], r['recency'], r['T'])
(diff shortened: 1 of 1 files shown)


lifetimes/estimation.py
```diff
@@ -96,14 +96,17 @@ def customer_lifetime_value(self, transaction_prediction_model, frequency, recen
         df['frequency'] = frequency
         df['recency'] = recency
         df['T'] = T
+        df['monetary_value'] = monetary_value
```
rhydomako: I agree that applying the CLV calculation per individual is super useful. To that end, I would suggest changing `conditional_expected_average_profit` to return individual estimates. Otherwise, there isn't really a reason to have `customer_lifetime_value` inside the Gamma-Gamma model. In fact, it might make a lot of sense to move `customer_lifetime_value` to `utils` to decouple it from a specific model altogether...

jackyckma:
In fact, in my own project I modified the `customer_lifetime_value()` such that it returns `discounted_monthly_cash_flows` instead of `sum(discounted_monthly_cash_flows)` so that I can work with individual estimates, as well as average or sum of those estimates. I feel that's more flexible to work this way but since it changes the output of the method significantly so I just didn't make such request.

BTW, I do feel it's more reasonable to use `conditional_expected_average_profit` for individual estimates too. Just don't have time to do the test recently.

rhydomako:
Personally, I think returning `discounted_monthly_cash_flows` instead of `sum(discounted_monthly_cash_flows)` and changing `conditional_expected_average_profit` to return individual estimates is the way to go. Even though these would be breaking changes, I think they would make the API much more flexible. I can tackle that part if it is outside the scope of your PR.

CamDavidsonPilon:
I'm okay merging this as is, and allowing future PRs to abstract pieces out. Thanks for the contribution @jackyckma :+1:

PR #51, opened 2016-02-14
Matplotlib tests

rhydomako:
This PR attempts to improve test coverage for the plotting routines by leveraging the pytest-mpl plugin. The idea here is for the tests to compare the matplotlib output to reference pngs and compare the RMS difference between the two plots (https://github.com/astrofrog/pytest-mpl).

This adds some additional complexity to the test framework because:
* an additional dependency
* reference plots have to be generated for any new plotting tests (not a big deal via `py.test --mpl-generate-path=tests/reference_plots`
* RMS values are pretty sensitive to things like text placement, which looks like it might vary between matplotlib releases (this issue is mitigated by tuning the reference tolerance)

But the benefits are:
* strong test coverage for the plotting routines (both that they produce output and that the result is close to the test reference)
* weak constraints on the `generate_data` routines -- the `plot_period_transactions` routine relies on the the generate_data code, which wasn't really getting called during testing before. This is a rather indirect way of testing that code, and it doesn't really verify that the routines are providing the correct output other than that the resultant plots are similar to what is expected. However, this should still be enough to detect gross regressions at test time on code that wasn't really being covered before.


lifetimes/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .estimation import BetaGeoFitter, ParetoNBDFitter, GammaGammaFitter
+from .estimation import BetaGeoFitter, ParetoNBDFitter, GammaGammaFitter, ModifiedBetaGeoFitter
 from .version import __version__
 
-__all__ = ['BetaGeoFitter', 'ParetoNBDFitter', 'GammaGammaFitter']
+__all__ = ['BetaGeoFitter', 'ParetoNBDFitter', 'GammaGammaFitter', 'ModifiedBetaGeoFitter']
```
CamDavidsonPilon: :+1: good call

CamDavidsonPilon:
I use a different matplotlibrc file locally, that has different styles (mostly default colours). How will these tests behave for that? 

(pulling locally to test now)

CamDavidsonPilon:
Hm, so I also think because my default DPI is much higher, I get errors like:
```
ValueError: operands could not be broadcast together with shapes (600,800,3) (2400,3300,3)
```

However, I think it's great that TravisCI can now test the images (and it's :green_apple:!). I'm okay forgoing locally testing (and when I do, I can change my default matplotlibrc) so that remote testing is more correct.

:+1: great additions, I didn't know about this lib until now!

rhydomako:
Hmm, a local matplotlibrc file would make this a bit more tricky.

I think you could generate your own local reference plots like:

    py.test --mpl-generate-path=tests/my_reference_plots

Then when you want to test the plotting routines you could just point to those reference plots:

    py.test --mpl --mpl-baseline-path tests/my_reference_plots tests/test_plotting.py

But conveniently, if you just run

    py.test

without the `--mpl` flag, the plotting tests will be run, but they'll only check that the code runs and won't compare the images.

PR #52, opened 2016-02-15
Fix matplotlib test config that broke travis-ci

rhydomako:
When adding  pytest-mpl tests (#51), I missed some travis-ci configurations to get the whole thing working. Specifically:
* had to change the matplotlib to a non-interactive backend so that it doesn't have to deal with DISPLAY problems
* explicitly include matplotlib in the travis test environment
* for some reason, the python 3.3 environment needs `nose` to be installed for some part of matplotlib to function properly...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
   - conda info -a
 
   # Should match requirements.txt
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
:+1: cool - I wonder why it was :green_apple: on the other PR?

PR #57, opened 2016-04-24
Add BG/BB discrete transactional model

benvandyke:
Hello Cameron, this PR adds the BG/BB model described by Fader and Hardie ([PDF](https://marketing.wharton.upenn.edu/files/?whdmsaction=public:main.file&fileID=344)) for modeling discrete time transactions where no death is observed. The model is useful for cases when repeat purchases occur at regular intervals such as in subscription businesses. This PR contains the model, some tests, and the dataset provided by Hardie on his website. 

BG/BB is also available in the BTYD R package, I believe with this addition lifetimes will offer Python implementations of all the models available in that package.


tests/test_estimation.py
```diff
@@ -6,10 +6,82 @@
...
+
+        bbtf = estimation.BetaGeoBetaBinomFitter()
+        bbtf.fit(
+            donations['frequency'],
+            donations['recency'],
+            donations['n'],
+            donations['n_custs'],
+        )
+        expected = np.array([1.204, 0.750, 0.657, 2.783])
+        npt.assert_array_almost_equal(expected, np.array(bbtf._unload_params('alpha','beta','gamma','delta')),
```
CamDavidsonPilon: awesome

lifetimes/estimation.py
```diff
@@ -36,6 +37,185 @@ def _print_params(self):
...
+
+    """
+    
+    def __init__(self, penalizer_coef=0.):
+        self.penalizer_coef = penalizer_coef
+
+    @staticmethod
+    def _loglikelihood(params, x, tx, T):
+
+        N = len(x)
```
CamDavidsonPilon: to avoid ambiguity, can you use `x.shape[0]`? (I think it is safe to assume only numpy arrays will be allowed in here.

lifetimes/estimation.py
```diff
@@ -36,6 +37,185 @@ def _print_params(self):
...
+        def _sum(x, tx, recency_T):
+
+            j = np.arange(recency_T+1)
+            return log(
+                np.sum(exp(special.betaln(alpha + x, beta + tx - x + j) - beta_ab +
+                              special.betaln(gamma + 1, delta + tx + j) - beta_gd)))
+
+        for i in np.arange(N):
+            indiv_loglike[i] = logaddexp(indiv_loglike[i],_sum(x[i], tx[i], recency_T[i]))
+
```
CamDavidsonPilon: instead of the for loop, can we just do?
```
indiv_loglike =  logaddexp(indiv_loglike,_sum(x, tx, recency_T))
```
maybe not because of the `_sum` function

lifetimes/estimation.py
```diff
@@ -36,6 +37,185 @@ def _print_params(self):
...
+    def fit(self, frequency, recency, n, n_custs, verbose=False):
+        """
+        Fit the BG/BB model.
+
+        Parameters:
+            frequency: Total periods with observed transaction
+            recency: Period of most recent transaction
+            n: Number of transaction opportunities
+            n_custs: Number of customers with frequency/recency/n pattern
```
CamDavidsonPilon: I'm not sure what this parameter means in this context

lifetimes/estimation.py
```diff
@@ -36,6 +37,185 @@ def _print_params(self):
...
+        params = self._unload_params('alpha', 'beta', 'gamma', 'delta')
+        alpha, beta, gamma, delta = params
+
+        x_counts = self.data.groupby('frequency')['n_custs'].sum()
+        x = asarray(x_counts.index)
+
+        p1 = special.binom(n, x) * exp(special.betaln(alpha + x, beta + n - x) - special.betaln(alpha, beta) +
+         special.betaln(gamma, delta + n) - special.betaln(gamma, delta))
+
+        for j in np.arange(len(x)):
```
CamDavidsonPilon: ditto above, can you avoid `len` here?

CamDavidsonPilon:
Thanks for this addition! A few points initially:

1. I haven't read the paper (yet), so I can't comment on the algorithm implementation yet
2. Really great tests! 🎉

lifetimes/estimation.py
```diff
@@ -36,6 +37,185 @@ def _print_params(self):
...
+        self.data = DataFrame(vconcat[frequency, recency, n, n_custs],
+                              columns=['frequency','recency','n','n_custs'])
+
+        return self
+
+    def conditional_expected_number_of_purchases_up_to_time(self, n_star):
+        """
+        Conditional expected purchases in future time period.
+
+        The  expected  number  of  future  transactions across the next n* transaction
```
CamDavidsonPilon: what is n_star? That variable name is not very human readable

benvandyke: That's the notation Fader and Hardie used for the conditional expectation. In the Pareto/NBD and BG/NBD papers they used _t_ instead. In my most recent commit I updated it to _t_ for consistency.

lifetimes/estimation.py
benvandyke: Removed both calls to `len` in latest commit.

lifetimes/estimation.py
benvandyke: I added some clarification in the doc string of what `n_custs` is used for.

lifetimes/estimation.py
benvandyke: I tried to remove the loop during initial development, but `_sum` is already summing over the `j` vector. I'll take another look and see if it can be cleaned up.

benvandyke:
@CamDavidsonPilon Hey Cameron, just wanted to check in with you on this. Have you had a chance to read Fader and Hardie's paper? Feel free to shoot me any questions on the model or implementation.

CamDavidsonPilon:
Just read it now - really useful model.

CamDavidsonPilon:
Overall, lgtm! I'll merge now

PR #59, opened 2016-06-06
Updated Utils.py

eshaankuls25:
calibration_summary_data = summary_data_from_transaction_data(calibration_transactions, customer_id_col, datetime_col,monetary_value_col,
                                                                  datetime_format, observation_period_end=calibration_period_end, freq=freq)

monetary_value_col is not added as an argument

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -48,7 +48,7 @@ def calibration_and_holdout_data(transactions, customer_id_col, datetime_col, ca
 
     # create calibration dataset
     calibration_transactions = transactions.ix[transactions[datetime_col] <= calibration_period_end]
-    calibration_summary_data = summary_data_from_transaction_data(calibration_transactions, customer_id_col, datetime_col,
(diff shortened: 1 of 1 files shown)


lifetimes/utils.py
```diff
@@ -48,7 +48,7 @@ def to_period(d):
 
     # create calibration dataset
     calibration_transactions = transactions.ix[transactions[datetime_col] <= calibration_period_end]
-    calibration_summary_data = summary_data_from_transaction_data(calibration_transactions, customer_id_col, datetime_col,
+    calibration_summary_data = summary_data_from_transaction_data(calibration_transactions, customer_id_col, datetime_col,monetary_value_col,
```
CamDavidsonPilon: does `monetary_value_col` need to be added the the function args too?

eshaankuls25: Actually,when I tried implementing the previous function, it was throwing some error.

PR #62, opened 2016-08-04
Added Python 3.5 to CI and explicit pypi trove classifiers

cfournie:
Adding this will allow libraries like [`caniusepython3`](https://pypi.python.org/pypi/caniusepython3) and [sites that rely on it](http://py3readiness.org/) to recognize that this library supports Python 3.


CamDavidsonPilon:
woohoo!

PR #63, opened 2016-08-18
Clarifying discount rate is monthly

kingfink:
This library has been super-useful, I was getting mixed-up on the discount rate and figured it might be helpful to be explicit in the documentation, and change the default to something a bit more reasonable.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![logo](http://i.imgur.com/7s3jqZM.png)
 
-#### Measuring users is hard. Lifetimes makes it easy. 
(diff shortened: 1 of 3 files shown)


CamDavidsonPilon:
good suggestion, thanks for the PR!

CamDavidsonPilon:
@kingfink there are some tests that need to be updated with this - do you mind adding those changes too?

kingfink:
@CamDavidsonPilon absolutely, just fixed the issues. Thanks for the review, and thanks again for open sourcing this tool!

PR #66, opened 2016-09-30
replacing to-be-deprecated sort to sort_values(by=) in the README

kevin-mercurio:
Working through the tutorial in Jupyter, it seems that the sort() method on dataframes is deprecated. Just replaced that text in the README.md with  sort_values(by= ...)


CamDavidsonPilon:
👍 nice, thanks for the contribution!

PR #70, opened 2016-10-07
Optimize Performance of Estimators

HHammond:
So for this PR has halved the amount of time that the test suite for estimators takes. Still looking at some more improvements.

cc @CamDavidsonPilon

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -59,27 +59,27 @@ class BetaGeoBetaBinomFitter(BaseFitter):
     @staticmethod
     def _loglikelihood(params, x, tx, T):
 
-        N = x.shape[0]
-
         alpha, beta, gamma, delta = params
 
         beta_ab = special.betaln(alpha, beta)
         beta_gd = special.betaln(gamma, delta)
 
         indiv_loglike = (special.betaln(alpha + x, beta + T - x) - beta_ab +
-                        special.betaln(gamma, delta + T) - beta_gd)
+                         special.betaln(gamma, delta + T) - beta_gd)
 
         recency_T = T - tx - 1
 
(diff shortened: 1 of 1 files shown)


lifetimes/estimation.py
```diff
@@ -71,15 +69,16 @@ def _loglikelihood(params, x, tx, T):
 
         recency_T = T - tx - 1
 
+        js = np.arange(max(recency_T) + 1)
+        @np.vectorize
         def _sum(x, tx, recency_T):
-
-            j = np.arange(recency_T+1)
+            j = js[:recency_T + 1]
```
HHammond: This was doing n^2 memory allocations. Now the values are stored in memory once which should cut the cost of computing loglikelihood by it's square root.

lifetimes/estimation.py
```diff
@@ -640,7 +639,8 @@ def probability_of_n_purchases_up_to_time(self, t, n):
 
         first_term = special.beta(a, b + n) / special.beta(a, b) * special.gamma(r + n) / special.gamma(r) / special.gamma(n + 1) * (alpha / (alpha + t)) ** r * (t / (alpha + t)) ** n
         if n > 0:
-            finite_sum = np.sum([special.gamma(r + j) / special.gamma(r) / special.gamma(j + 1) * (t / (alpha + t)) ** j for j in range(0, n)])
+            j = np.arange(0, n)
+            finite_sum = (special.gamma(r + j) / special.gamma(r) / special.gamma(j + 1) * (t / (alpha + t)) ** j).sum()
```
CamDavidsonPilon: yesss so obvious in retrospect

HHammond:
@CamDavidsonPilon this is good IMO, I don't see any more big wins.

PR #73, opened 2016-10-10
Change fit method

CamDavidsonPilon:
- `penalizer_coef` was doing nothing  in BetaGeoBetaBinomFitter, so I fixed that
- removed `Powell`  as a fitter. It gave the same results as NM, and was significantly slower. 
- added a warm-start to the fitters: if the previous fitter converges, use the output as an initial input. 
- instead of just running `iterative_fitting` times, we will run until `iterative_fitting` successful convergences. This, while appears to slow down the entire _fit routine, combined with the above to changes, makes the `test_estimator.py` test suite 50% faster!

EDIT:
 - I'm also reducing the tolerance from `1e-6` to `1e-4`. I never care about such precision, so I'd rather have a faster run time than more precision I might never need. 
 - Also bumping the version, but I'll wait for your changes in https://github.com/CamDavidsonPilon/lifetimes/pull/70 before I upload to PyPI

cc @hhammond for review please 

#### branch
```
1 loop, best of 3: 2min 21s per loop
OrderedDict([('r', 0.51888231072581714), ('alpha', 21.222120157343589), ('a', 2.3424582901483939), ('b', 4.3620516233819249)])
3348411.23857
```

#### master
```
1 loop, best of 1: 6min 49s per loop
OrderedDict([('r', 0.51868238084360141), ('alpha', 21.209590272922487), ('a', 2.3439241385746037), ('b', 4.3625653897493084)])
3348758.63896
```

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -261,16 +261,15 @@ class GammaGammaFitter(BaseFitter):
         p, q, v = self._unload_params('p', 'q', 'v')
         return (((q - 1) / (p * x + q - 1)) * (v * p / (q - 1))) + (p * x / (p * x + q - 1)) * m
 
-    def fit(self, frequency, monetary_value, iterative_fitting=5, initial_params=None, verbose=False):
+    def fit(self, frequency, monetary_value, iterative_fitting=4, initial_params=None, verbose=False):
         """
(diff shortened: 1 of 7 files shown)


HHammond:
LGTM

PR #77, opened 2016-10-31
Check input values of GammaGammaFitter.fit

yanirs:
Add check to ensure that monetary values passed into `GammaGammaFitter.fit` are all positive.

Background: I spent some time figuring out why I was getting negative value predictions from `GammaGammaFitter`, and it turned out that I was passing in customers with zero mean value. Checking it explicitly would prevent other users from making the same mistake.

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -109,7 +109,7 @@ class BetaGeoBetaBinomFitter(BaseFitter):
         frequency = asarray(frequency)
         recency = asarray(recency)
         n = asarray(n)
-        n_custs= asarray(n_custs)
+        n_custs = asarray(n_custs)
         _check_inputs(frequency, recency, n)
 
         params, self._negative_log_likelihood_ = _fit(self._negative_log_likelihood,
@@ -259,9 +259,9 @@ class GammaGammaFitter(BaseFitter):
         """
         This method computes the conditional expectation of the average profit per transaction
         for a group of one or more customers.
-            x: a vector containing the customers' frequencies. Defaults to the whole set of
+            frequency: a vector containing the customers' frequencies. Defaults to the whole set of
                 frequencies used for fitting the model.
-            m: a vector containing the customers' monetary values. Defaults to the whole set of
+            monetary_value: a vector containing the customers' monetary values. Defaults to the whole set of
                 monetary values used for fitting the model.
 
         Returns:
@@ -287,6 +287,8 @@ class GammaGammaFitter(BaseFitter):
         Returns:
             self, fitted and with parameters estimated
         """
+        _check_inputs(frequency, monetary_value=monetary_value)
+
         params, self._negative_log_likelihood_ = _fit(self._negative_log_likelihood,
                                                       [frequency, monetary_value, self.penalizer_coef],
                                                       iterative_fitting,
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -261,24 +261,16 @@ def _scale_time(age):
     return 10. / age.max()
 
 
-def _check_inputs(frequency, recency, T):
-
-    def check_recency_is_less_than_T(recency, T):
-        if np.any(recency > T):
-            raise ValueError("""Some values in recency vector are larger than T vector. This is impossible according to the model.""")
-
-    def check_frequency_of_zero_implies_recency_of_zero(frequency, recency):
-        ix = frequency == 0
-        if np.any(recency[ix] != 0):
-            raise ValueError("""There exist non-zero recency values when frequency is zero. This is impossible according to the model.""")
-
-    def check_all_frequency_values_are_integer_values(frequency):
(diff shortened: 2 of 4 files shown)


lifetimes/estimation.py
```diff
@@ -259,9 +259,9 @@ def conditional_expected_average_profit(self, frequency=None, monetary_value=Non
         """
         This method computes the conditional expectation of the average profit per transaction
         for a group of one or more customers.
-            x: a vector containing the customers' frequencies. Defaults to the whole set of
+            frequency: a vector containing the customers' frequencies. Defaults to the whole set of
```
yanirs: Unrelated fix to make the docstring match the actual argument names.

lifetimes/utils.py
```diff
@@ -261,24 +261,16 @@ def _scale_time(age):
     return 10. / age.max()
 
 
-def _check_inputs(frequency, recency, T):
-
-    def check_recency_is_less_than_T(recency, T):
```
yanirs: Not sure why these were all nested functions... As far as I can tell, it works just as well without the functions, and the error strings serve as documentation of the error conditions (in case documentation was the original intent).

tests/test_utils.py
```diff
@@ -260,22 +260,30 @@ def test_calculate_alive_path(example_transaction_data, example_summary_data, fi
 
 
 def test_check_inputs():
-    freq, recency, T = np.array([0,1,2]), np.array([0, 1, 10]), np.array([5, 6, 15])
-    assert utils._check_inputs(freq, recency, T) is None
+    frequency = np.array([0, 1, 2])
```
yanirs: Renamed `freq` to `frequency` for consistency and added testing of `monetary_value`

tests/test_utils.py
```diff
@@ -260,22 +260,30 @@ def test_calculate_alive_path(example_transaction_data, example_summary_data, fi
...
+        utils._check_inputs(frequency, bad_recency, T)
 
     with pytest.raises(ValueError):
         bad_freq = np.array([0, 0.5, 2])
         utils._check_inputs(bad_freq, recency, T)
 
+    with pytest.raises(ValueError):
+        bad_monetary_value = monetary_value.copy()
+        bad_monetary_value[0] = 0
+        utils._check_inputs(frequency, recency, T, bad_monetary_value)
```
CamDavidsonPilon: 👍

tests/test_estimation.py
```diff
@@ -10,6 +10,9 @@
 
 cdnow_customers = load_cdnow()
 cdnow_customers_with_monetary_value = load_summary_data_with_monetary_value()
+returning_cdnow_customers_with_monetary_value = cdnow_customers_with_monetary_value[
+    cdnow_customers_with_monetary_value['frequency'] > 0
+]
```
CamDavidsonPilon: why do we need to do this?

yanirs: For `TestGammaGammaFitter.test_params_out_is_close_to_Hardie_paper`. I can move this variable to that test if you prefer. The problem is that `cdnow_customers_with_monetary_value` contains many customers with monetary value of zero, so the new check fails. Fitting the model only on returning customers matches the example from the readme (https://github.com/CamDavidsonPilon/lifetimes#estimating-customers-lifetime-value).

CamDavidsonPilon: yea I think it's better to move this into the test, to make it explicit. I wonder why there are zero money values in the source data 😕

yanirs: I moved it into the test. The zero values come from Hardie's spreadsheet -- it's because they calculate the mean value of *repeat* transactions. Not sure why the model can't just use the mean transaction value. The answer may be in the original paper, which I haven't read yet (can't find it in http://www.brucehardie.com/notes/025/gamma_gamma.pdf)

CamDavidsonPilon:
👍 good one, ty @yanirs!

lifetimes/utils.py
```diff
@@ -261,24 +261,16 @@ def _scale_time(age):
...
+def _check_inputs(frequency, recency=None, T=None, monetary_value=None):
+    if recency is not None:
+        if T is not None and np.any(recency > T):
+            raise ValueError("Some values in recency vector are larger than T vector.")
+        if np.any(recency[frequency == 0] != 0):
+            raise ValueError("There exist non-zero recency values when frequency is zero.")
+    if np.sum((frequency - frequency.astype(int)) ** 2) != 0:
+        raise ValueError("There exist non-integer values in the frequency vector.")
+    if monetary_value is not None and np.any(monetary_value <= 0):
+        raise ValueError("There exist non-positive values in the monetary_value vector.")
```
harrisong: @yanirs thanks for the great PR! Just want to know the ramification if the `monetary_value == 0`. In the PR, looks like negative values will contribute to negative CLV. However, is it the same case for the zero values? This is important to our team as it will distort the overall average CLV across all the customers after we pass through the `customer_lifetime_value` method below. Any help is appreciated, thanks!

Issue #33, opened 2015-12-30
_print_params is a bad name


CamDavidsonPilon:
https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/estimation.py#L33

doesn't even print!

PR #78, opened 2016-10-31
Add tests for BaseFitter

yanirs:
I felt bad about the tiny decrease in coverage in https://github.com/CamDavidsonPilon/lifetimes/pull/77 so I added some tests for `BaseFitter` :)

Also fixed https://github.com/CamDavidsonPilon/lifetimes/issues/33 while I was at it (just got rid of the method in favour of a one-liner).


CamDavidsonPilon:
👍 great!

PR #79, opened 2016-11-01
Improve documentation of GammaGammaFitter

yanirs:
I think this makes it a bit easier to understand. It's actually a pretty simple model :slightly_smiling_face:

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -295,7 +295,6 @@ class GammaGammaFitter(BaseFitter):
             frequency: the frequency vector of customers' purchases (denoted x in literature).
             monetary_value: the monetary value vector of customer's purchases (denoted m in literature).
             iterative_fitting: perform iterative_fitting fits over random/warm-started initial params.
-                 This model is not very stable so we suggest >10 for best estimates evaluation.
             initial_params: set initial params for the iterative fitter.
             verbose: set to true to print out convergence diagnostics.
             tol: tolerance for termination of the function minimization process.
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -32,6 +32,7 @@ class BaseFitter(object):
         return [self.params_[x] for x in args]
 
 
+
 class BetaGeoBetaBinomFitter(BaseFitter):
     """
     Also known as the Beta-Geometric/Beta-Binomial Model Model [1].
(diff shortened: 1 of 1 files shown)


lifetimes/estimation.py
```diff
@@ -262,10 +277,15 @@ def conditional_expected_average_profit(self, frequency=None, monetary_value=Non
...
-        x = self.data['frequency'] if frequency is None else frequency
+        if monetary_value is None:
+            monetary_value = self.data['monetary_value']
+        if frequency is None:
+            frequency = self.data['frequency']
         p, q, v = self._unload_params('p', 'q', 'v')
-        return (((q - 1) / (p * x + q - 1)) * (v * p / (q - 1))) + (p * x / (p * x + q - 1)) * m
+        # The expected average profit is a weighted average of individual monetary value and the population mean.
+        individual_weight = p * frequency / (p * frequency + q - 1)
+        population_mean = v * p / (q - 1)
```
yanirs: This is actually a bit confusing because the meaning of `v` in the paper is different. In the paper, `v` is called &gamma; (see equation 4 in http://www.brucehardie.com/papers/rfm_clv_2005-02-16.pdf), which is also a less-than-ideal choice given the many meanings of gamma...

In any case, I think that this change makes it easier to see how the prediction is made.

lifetimes/estimation.py
```diff
@@ -278,6 +298,7 @@ def fit(self, frequency, monetary_value, iterative_fitting=4, initial_params=Non
                  This model is not very stable so we suggest >10 for best estimates evaluation.
             initial_params: set initial params for the iterative fitter.
             verbose: set to true to print out convergence diagnostics.
+            tol: tolerance for termination of the function minimization process.
```
yanirs: Have you considered passing some of these arguments to `__init__` rather than to `fit`? That'd be more consistent with sklearn conventions.

lifetimes/estimation.py
```diff
@@ -278,6 +298,7 @@ def fit(self, frequency, monetary_value, iterative_fitting=4, initial_params=Non
                  This model is not very stable so we suggest >10 for best estimates evaluation.
```
CamDavidsonPilon: bah, good question - thanks for calling me out. I think it was for perf reasons initially

yanirs: Depending on what you mean by stability, you may not need many iterations. At least on my dataset, removing monetary values of zero made a big difference to stability when measuring the MAE of the predictions. Before dropping zeros, multiple runs on the same data yielded highly variable results. Which dataset did you use and how did you measure stability?

lifetimes/estimation.py
CamDavidsonPilon: I like that suggestion, let's make an issue and move ahead with that 👍

lifetimes/estimation.py
CamDavidsonPilon: 👌 good additions - very clear variable names

lifetimes/estimation.py
CamDavidsonPilon: The cdnow_customers_transactions data - I wonder if the zeros were the cause... 
(the original docs were included in the initial PR: https://github.com/CamDavidsonPilon/lifetimes/pull/22)

CamDavidsonPilon: 👍 cool with that

CamDavidsonPilon:
lgtm, I'll merge after your doc-string change

PR #82, opened 2016-11-26
changing how we use do the penalizer term to be more straightforward

CamDavidsonPilon:
This changes how `penalizer_coef` behaves (users may need to re-cross validate their work). See my explaination here: https://dataorigami.net/blogs/napkin-folding/a-real-life-mistake-i-made-about-penalizer-terms

cc @yanirs for review + thoughts please

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -81,7 +81,7 @@ class BetaGeoBetaBinomFitter(BaseFitter):
 
     @staticmethod
     def _negative_log_likelihood(params, frequency, recency, n, n_custs, penalizer_coef=0):
-        penalizer_term = penalizer_coef * log(params).sum()
+        penalizer_term = penalizer_coef * sum(params)
         return -np.sum(BetaGeoBetaBinomFitter._loglikelihood(params, frequency, recency, n) * n_custs) + penalizer_term
 
     def fit(self, frequency, recency, n, n_custs, verbose=False, tol=1e-4, iterative_fitting=1):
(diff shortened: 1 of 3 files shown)


yanirs:
@CamDavidsonPilon An interesting mistake indeed :slightly_smiling_face: 

I read the post and it all makes sense. However, what do you think about supporting both L1 and L2 penalty terms (like an elastic net)?

It'd also be good to add a note in the readme that this is a possibly-breaking change.

lifetimes/estimation.py
```diff
@@ -81,7 +81,7 @@ def _sum(x, tx, recency_T):
 
     @staticmethod
     def _negative_log_likelihood(params, frequency, recency, n, n_custs, penalizer_coef=0):
-        penalizer_term = penalizer_coef * log(params).sum()
+        penalizer_term = penalizer_coef * sum(params)
```
yanirs: Won't `params.sum()` be faster in this case?

CamDavidsonPilon: params is sometimes a Python list, so no builtin sum. The list has maximum 4 elements, and doing some offline `timeit` tests shows that Python's `sum` is actually really fast

CamDavidsonPilon:
> I read the post and it all makes sense. However, what do you think about supporting both L1 and L2 penalty terms (like an elastic net)?

I'm not sure - it complicates the API, and I don't think many users will use both parameters. For the "different scales" argument outlined in the blog post, I think the L2 penalty term could do more harm than good (at least, wouldn't add anything that the L1 term already provided)

CamDavidsonPilon:
I'm not sure what's up with circle, test locally and all 🍏

yanirs:
@CamDavidsonPilon Sounds good. I was also unsure it's worth complicating the API.

Maybe the problem with the build is a memory issue? It doesn't seem related to the change in any case...

PR #86, opened 2016-12-14
Update lifetimes.utils.find_first_transactions to sort by datetime_col before finding first transaction

aashanand:


CamDavidsonPilon:
👍 looks good. I wonder if any tests will fail because of this (actually: I don't think Travis is working atm). 

Also: can you add a test to `test_utils.py` too? This seems like something that can be missed in future refactors.

CamDavidsonPilon:
Can you email me at cam.davidson.pilon@gmail.com? I want to chat with you about how you are using this at earthled.com

aashanand:
Thanks - I'll get you an update on this tomorrow.

PR #88, opened 2017-01-01
I didn't like how summary_from_transaction was squashing orders from …

CamDavidsonPilon:
…the same period to a single order. That didn't really make sense: it's reasonable to have <1 rate of purchases.


rhydomako:
I should note that accounting for interval-censored transaction data introduces some subtle model changes: http://www.brucehardie.com/notes/011/Pareto_NBD_interval_censored_2010-08-07.pdf

However, I don't have any actual suggestions about a good way to deal with this.

PR #89, opened 2017-01-03
Pin the pytest-mpl version to 0.5

rhydomako:
Looks like the pytest-mpl plugin pushed some releases in late Nov, which appears to be what is breaking the travis build.

I couldn't figure out what the problem is with pytest-mpl v0.7, as it is one of those annoying problems that don't occur locally. So a simple fix for the moment is to pin pytest-mpl to v0.5, which should at least get the CI and coverage going again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - source activate test-environment
   - python setup.py install
   - pip install coveralls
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
I love seeing all this green! Thanks @rhydomako for tracking this down!

PR #98, opened 2017-05-25
update utils for python3 and modern pandas

aprotopopov:


lifetimes/utils.py
```diff
@@ -110,7 +110,11 @@ def find_first_transactions(transactions, customer_id_col, datetime_col, monetar
     if monetary_value_col:
         select_columns.append(monetary_value_col)
 
-    transactions = transactions[select_columns].sort(select_columns).copy()
+    # in python 3.6 with pandas 0.20.1 sort raise error, sort_values was added in 0.17.0
+    if int(pd.__version__.replace('.', '')) >= 170:
+        transactions = transactions[select_columns].sort_values(select_columns).copy()
+    else:
+        transactions = transactions[select_columns].sort(select_columns).copy()
```
CamDavidsonPilon: hm, in the requirements I explicitly ask for 0.19. Should we bump this to 0.20 and make this only sort_values?

aprotopopov: Ok, I missed that. So let's leave sort_values and delete all other rows.

aprotopopov:
I deleted all staff about sort and pandas versions.

aprotopopov:
Also did some tricks with `find_first_transactions` to find first transactions, for now it's more clear IMHO. Tried to improve performance but without success yet.

aprotopopov:
Another little change, now all pytests passed successfully for python2 and python3.

tests/test_estimation.py
```diff
@@ -91,7 +91,7 @@ def test_conditional_expectation_returns_same_value_as_Hardie_excel_sheet(self):
         expected = 12884.2 # Sum of column F Exp Tot
         npt.assert_almost_equal(expected, pred_purchases.sum(), decimal=0)
 
-    def test_expected_purchases_in_n_periods_returns_same_value_as_Hardie_excel_sheet(selfself):
```
CamDavidsonPilon: silly me

CamDavidsonPilon:
All non-matplotlib tests pass (I really need to fix those, or just get rid of them), so it's 🍏 to me

lifetimes/utils.py
```diff
@@ -110,31 +110,25 @@ def find_first_transactions(transactions, customer_id_col, datetime_col, monetar
...
+        # drop duplicated rows
+        period_transactions = transactions.drop_duplicates()
 
     # initialize a new column where we will indicate which are the first transactions
-    period_transactions['first'] = False
-    # find all of the initial transactions and store as an index
-    first_transactions = period_transactions.groupby(customer_id_col, sort=True, as_index=False).head(1).index
-    # mark the initial transactions as True
-    period_transactions.loc[first_transactions, 'first'] = True
+    period_transactions['first'] = ~(period_transactions.duplicated(customer_id_col))
```
CamDavidsonPilon: what does `period_transactions.duplicated(customer_id_col)` do?

aprotopopov: It's indicate if row is duplicated or not. First argument is subset for which it's looking for. https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.duplicated.html. May be it's coming more unclear with that :). If so we could revert to previous behavior

CamDavidsonPilon: Hm, but aren't we dropping any duplicates in 
> transactions.drop_duplicates()

above - so there would be no duplicates?

(I'm obviously missing something)

aprotopopov: It's all defaults.. [drop_duplicates](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.drop_duplicates.html) by default drops by all columns in that case date and id, while then we looking only for duplicates by 'id' which we specify as first argument. Is that clearer now?

CamDavidsonPilon: ok ok, makes sense, but then why does it necessarily satisfy:
```
# mark the initial transactions as True
```

aprotopopov: Because when `duplicated` saw row and didn't see it yet it marks as False. Then we doing `~` as opposite. I did that because I thought that it'll be faster... But benchmarks show no speed improvement. May be it's better to revert to previous with so many aspects of `duplicated` and `drop_duplicates`.

CamDavidsonPilon: I've tried optimising this code, and it's not trivial to get much faster =)

If there is not perf gain, I feel the former is more readable (imo) - so if you don't mind reverting back I would be happy with that.

aprotopopov: Ok, I'll revert back on Monday and if I find better performance I'll open new PR.

aprotopopov:
I reverted back changes about find_first_transactions and squashed everything to one commit.

aprotopopov:
Added cumulative and incremental plots with [commit](https://github.com/CamDavidsonPilon/lifetimes/pull/98/commits/30fb2268624555fe5773de17f77c2011e55c8aaa). If you everything is OK we could close [issue](https://github.com/CamDavidsonPilon/lifetimes/issues/99). I'm also plan to add some tests for utils and plotting for that function. And optionally update main markdown with new plots.

CamDavidsonPilon:
Neat - can you show off some figures in this PR too please?

aprotopopov:
```
%matplotlib inline
from lifetimes import BetaGeoFitter
from lifetimes.utils import summary_data_from_transaction_data
from lifetimes.plotting import plot_cumulative_transactions, plot_incremental_transactions
import pandas as pd
import matplotlib.pyplot as plt
plt.rcParams['figure.figsize'] = (15, 10)

transactions  = load_transaction_data()
transactions['date'] = pd.to_datetime(transactions['date'])

t = 50
observation_period_end =  transactions.date.min() + pd.Timedelta(days=t)
freq = 'D'

transactions_summary = summary_data_from_transaction_data(transactions, 'id', 'date', 
                                                          observation_period_end=observation_period_end, freq=freq)

bgf = BetaGeoFitter(penalizer_coef=0.005)

bgf.fit(transactions_summary['frequency'], transactions_summary['recency'], transactions_summary['T'])

plt.figure()
plot_cumulative_transactions(bgf, transactions, 'date', 'id', 2*t, t,
                             datetime_format=None, freq=freq, set_index_date=False, grid=True)

plt.figure()
plot_incremental_transactions(bgf, transactions, 'date', 'id', 2*t, t,
                             datetime_format=None, freq=freq, set_index_date=False, grid=True)
```
image

image

CamDavidsonPilon:
lgtm, thanks!

PR #104, opened 2017-06-07
loading and saving models with dill package

aprotopopov:
I'm not sure about workaround for saving and deleting permissions, so test could failed for that reason. Needed [dill](https://github.com/uqfoundation/dill) package for saving, loading and for test.

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -6,6 +6,7 @@ import pandas as pd
 from numpy import log, exp, logaddexp, asarray, any as npany, c_ as vconcat,\
     isinf, isnan, ones_like
 from pandas import DataFrame
+import dill
 
 from scipy.special import gammaln, hyp2f1, beta, gamma, betaln, binom
 from scipy import misc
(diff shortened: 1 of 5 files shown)


lifetimes/estimation.py
```diff
@@ -31,6 +31,16 @@ def _unload_params(self, *args):
             raise ValueError("Model has not been fit yet. Please call the .fit method first.")
         return [self.params_[x] for x in args]
 
+    def save_model(self, path):
+        import dill
```
CamDavidsonPilon: What are your thoughts on adding this to the `reqiurements.txt`?

aprotopopov: ok, then `import dill` could be placed on top of the file.

aprotopopov:
Changed requirements to include `dill` package by default

CamDavidsonPilon:
For tests to pass, we also need to update `.tavis.yml` line 55

aprotopopov:
Yeah, all checks passed for the first time.

CamDavidsonPilon:
Thanks for the contribution!

PR #105, opened 2017-06-08
Fix loading model for __dict__ attribute

aprotopopov:
Now all attributes should be available [example](https://stackoverflow.com/a/25128214/4542359). Feel free to add new tests or ideas how to test saving/loading models

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 autopep8:
 	autopep8 --ignore E501,E241,W690 --in-place --recursive --aggressive lifetimes/
 
+# F401=imported but unused, E501=line too long
 lint:
-	flake8 lifetimes
+	flake8 --ignore F401,E501 lifetimes
 
 autolint: autopep8 lint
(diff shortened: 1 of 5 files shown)


aprotopopov:
Also added ability to add custom index for `fit` methods. It could be useful for saving model and loading it, when index could be your custom id's.

aprotopopov:
And one more commit to whether save data from fitter or not, because in my case it's about 15mln and then dill object is about 750Mb... Mostly from `data` field

aprotopopov:
Autopep8 was done. Also add F401 and E501 for Makefile for `flake8` to pass `make autolint` command.

CamDavidsonPilon:
In the `fit` methods, can you add docs for what the `index` kwarg does / should be?

aprotopopov:
Yeah, sure. Forgot about that. Is it clearer now?

CamDavidsonPilon:
lgtm, thanks!

PR #106, opened 2017-06-16
Add dill for required packages for setup.py 

aprotopopov:


CamDavidsonPilon:
👌 right - makes sense

PR #108, opened 2017-06-23
Change fmin optimizer to more general minimize for _fit method

aprotopopov:
Also thinking about add warning where there is no convergence.


CamDavidsonPilon:
👍 nice I like this alot

CamDavidsonPilon:
scipy's minimizers should display a warning if convergence fails, so I would leave that to them.

PR #110, opened 2017-06-26
fix README

aprotopopov:
Update some not working code from readme. Changed output for BetaGeoFitter (order of parameters). Changes `print` with braces to be compatible with python 2.7.x and python 3.x.
P.S. What you think about documentation with [Sphinx](http://www.sphinx-doc.org/en/stable/) with [readthedocs theme](https://github.com/rtfd/sphinx_rtd_theme)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,16 +32,16 @@ As emphasized by P. Fader and B. Hardie, understanding and acting on customer li
 
     pip install lifetimes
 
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
:+1: thanks for looking after this. All changes I like. 

w.r.t Sphinx - yes, it's just a lot of work to write the docs out initially. I can maybe plan a sprint to get it done.

PR #111, opened 2017-06-29
changed docstring for conditional_probability_alive for MBG/NBD

aprotopopov:
Small correction for docstring `conditional_probability_alive` for MBG/NBD.


CamDavidsonPilon:
👍

Issue #25, opened 2015-09-10
Pareto_NBD overflow


Ozzerov:
When we have customers with frequency > 300 _A_0 function returns 0, because both denominators (max_alpha_beta + rec) *\* r_s_x and (max_alpha_beta + age) *\* r_s_x return Inf.
But in _negative_log_likelihood we only need log of that value, which can be achieved by logsumexp function from scipy.misk.
Suppose we have to calculate log(a/A^r - b/B^r) when r is so large that A^r and B^r overflows to Inf.
In that case we can do the following: log(A_0) = log(a/A^r - b/B^r) = log(a_B^r - b_A^r) - r_log(A_B) = 
log(exp(a + r_log(B)) - exp(b + r_log(A))) - r_log(A_B)
now we can use logsumexp for first term to get correct result.

Same problem occurs in calculating conditional_probability_alive:
(alpha + T) *\* (r + x) \* A_0 overflows to Inf \* 0 and returns NaN
In contrast to _A_0 where we need to be fast for optimization routines, usually conditional_probability_alive executes just once, and we can avoid overflow by using mpmath from sympy which allows precise calculations without overflowing.

And i would also recommend to add an option for differential evolution solver from scipy.optimize  

Adding some functions here:

```
def _log_a_0(r, alpha, s, beta, freq, rec, age):
    min_ab, max_ab, t = (alpha, beta, r + freq) if alpha < beta else (beta, alpha, s + 1)
    abs_ab = max_ab - min_ab
    rsx = r + s + freq
    p_1, q_1 = hyp2f1(rsx, t, rsx + 1., abs_ab / (max_ab + rec)), (max_ab + rec)
    p_2, q_2 = hyp2f1(rsx, t, rsx + 1., abs_ab / (max_ab + age)), (max_ab + age)
    sign = ones(len(freq))
    return logsumexp([log(p_1) + rsx * log(q_2), log(p_2) + rsx * log(q_1)],
        axis=0, b=[sign, -sign]) - rsx * log(q_1 * q_2)
```

import sympy.mpmath as mp

```
def p_alive_precise(self, freq, rec, age):
    freq, rec, age = check_inputs(freq, rec, age)
    r, alpha, s, beta = self.pars
    log_a_0 = self._log_a_0(r, alpha, s, beta, freq, rec, age)

    def precise(freq_, age_, log_a_0_):
        return float(mp.power(alpha + age_, r + freq_) * mp.exp(log_a_0_))

[...]

PR #113, opened 2017-07-05
BUG: fix for some RuntimeWarnings with BetaGeoFitter

AdriMarteau:
When I was using the `BetaGeoFitter` class on some data I ran into a couple of `RuntimeWarning`s due to negative logarithms or overflowed exponentials (like #25) especially with very high value for `frequency`. 

I have modified 3 functions to avoid those warnings:
- `_negative_log_likelihood`
- `conditional_expected_number_of_purchases_up_to_time`
- `conditional_probability_alive` 

I have also included a test for testing that no overflow happens in `conditional_probability_alive`

### DIFF
--- a/lifetimes/estimation.py
+++ b/lifetimes/estimation.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 import numpy as np
 import pandas as pd
 from numpy import log, exp, logaddexp, asarray, any as npany, c_ as vconcat,\
-    isinf, isnan, ones_like
+    isinf, isnan, ones_like, where
 from pandas import DataFrame
 import dill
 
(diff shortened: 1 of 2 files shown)


tests/test_estimation.py
```diff
@@ -513,6 +513,20 @@ def test_fit_with_index(self, cdnow_customers):
         )
         assert (bgf.data.index == index).all() == False
 
+    def test_no_runtime_warnings_high_frequency(self, cdnow_customers):
```
CamDavidsonPilon: 👍 nice test

CamDavidsonPilon:
This is great, thanks for this. Merging now

lifetimes/estimation.py
```diff
@@ -699,8 +708,16 @@ def conditional_probability_alive(self, frequency, recency, T):
         Returns: a scalar
 
         """
+        ln_exp_max = 300.0
+
         r, alpha, a, b = self._unload_params('r', 'alpha', 'a', 'b')
-        return 1. / (1 + (frequency > 0) * (a / (b + frequency - 1)) * ((alpha + T) / (alpha + recency)) ** (r + frequency))
+        log_div = (r + frequency) * log(
```
statwonk: Nice fix @AdriMarteau! I was reading through the associated paper and wondering why the `log` were necessary. :-D

Issue #67, opened 2016-10-06
observation_period_end should default to the max transaction time


yanirs:
In `utils.calibration_and_holdout_data(transactions, customer_id_col, datetime_col, calibration_period_end, observation_period_end=datetime.today(), freq='D', datetime_format=None)`: `datetime.today()` is set once when the file is first read. This can be confusing when dealing with historical data, as the holdout column and evaluation results become dependent on the time the process is run.

It'd be better to default to `None` and then in the function body do:

``` python
if observation_period_end is None:
    observation_period_end = transactions[datetime_col].max()
```

yanirs:
Actually, this applies to all functions that accept `observation_period_end` as an argument.

CamDavidsonPilon:
Hm, I wonder about this. 

> holdout column and evaluation results become dependent on the time the process is run.

that's true, which isn't ideal, but I don't know if have the max observation as the period end is the best replacement. Interesting you brought this up.

yanirs:
I hit this issue when I was changing `plot_calibration_purchases_vs_holdout_purchases` to calculate error metrics rather than just plot them (something that may be worth contributing back, but that's a separate issue). I use the X last days (`holdout_days`) of the dataset as a holdout set, so I thought it'd be enough to pass in the max date minus `holdout_days` as `calibration_period_end` and ignore the `observation_period_end` argument. However, when calling `plot_calibration_purchases_vs_holdout_purchases`, I found it unintuitive that `summary.iloc[0]['duration_holdout']` (on https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/plotting.py#L59) is not equal to `holdout_days`. This affected the results when I replaced `summary.iloc[0]['duration_holdout']` with `holdout_days` in my code (which seemed like a sensible change, at least to me).

In general, I find it confusing that you get different results today and tomorrow if you run the example from the readme and omit the `observation_period_end` parameter:
```python
summary_cal_holdout = calibration_and_holdout_data(transaction_data, 'id', 'date', period_end='2014-09-01') 
bgf.fit(summary_cal_holdout['frequency_cal'], summary_cal_holdout['recency_cal'], summary_cal_holdout['T_cal'])
plot_calibration_purchases_vs_holdout_purchases(bgf, summary_cal_holdout)
```

In my opinion, the default value of `datetime.today()` is unlikely to be what users want (especially since it keeps changing). Do you have an idea for a better replacement than the max observation time?

[...]

PR #116, opened 2017-07-13
Small safety check on inputs to fitters.

statwonk:
Still a bit of a python noob, so forgive me if this is a bit pedantic --- getting up to speed with `lifetimes` and python. :)

Similar to the issue here, https://github.com/CamDavidsonPilon/lifetimes/issues/67 `observation_period_end` was confusing to me (until I read its definition 😛 !). I kept fumbling with it and hitting a deep `numpy` error caused by div by 0 [here](https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/utils.py#L290).

This check would have alerted me, that's all. :) Hoping to spend some time learning BTYD over the next couple of months!

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -298,6 +298,8 @@ def _check_inputs(frequency, recency=None, T=None, monetary_value=None):
             raise ValueError("There exist non-zero recency values when frequency is zero.")
         if np.any(recency < 0):
             raise ValueError("There exist negative recency (ex: last order set before first order)")
+        if any(len(x) == 0 for x in [recency, frequency, T]):
+            raise ValueError("There exists a zero length vector in one of frequency, recency or T.")
     if np.sum((frequency - frequency.astype(int)) ** 2) != 0:
         raise ValueError("There exist non-integer values in the frequency vector.")
     if monetary_value is not None and np.any(monetary_value <= 0):


CamDavidsonPilon:
Hey @statwonk! Hm, if you are hitting the error here: https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/utils.py#L290, then is your entire `T` column all zeros? I don't see how that could be 0 otherwise.

statwonk:
Hey @CamDavidsonPilon! The gist is that I was using `lifetimes` in an engineering context where the series was a span of two hours contained within a single day. I set the `observation_period_end` simply to the date like `2017-07-13` without realizing it'd truncate. I'm not sure how many people this fix'll help, but I figured it couldn't hurt and wanted to dip my toe into contributing some python code. :)

Hopefully more to come!

CamDavidsonPilon:
hm, so aside from being confused by `observation_period_end`, was the problem also that the default `freq` was in days? (and it sounds like all your data was from the current day, hence `max(T) == 0`).

statwonk:
I went back and tried to create a minimally reproducible example. I found that `age` was an empty `numpy` array rather than 0: `array([], dtype=float64)`.

Here's the example,

data: https://gist.github.com/statwonk/61ad1f6fff454dcd8a11b49632701a17
script (run from `/lifetimes/lifetimes/`): https://gist.github.com/statwonk/b93947b101308c0d2d94ade2216ab506

Here's the stacktrace,

```
2017-07-19 01:02:03.601827 # <--- min time
2017-07-19 17:19:08.323395 # <--- max time
```

```python
Traceback (most recent call last):
  File "inserter_p_alive.py", line 19, in <module>
    bgf.fit(summary['frequency'], summary['recency'], summary['T'])
  File "/Users/statwonk/Library/Python/2.7/lib/python/site-packages/lifetimes/fitters/beta_geo_fitter.py", line 79, in fit
    self._scale = _scale_time(T)
  File "/Users/statwonk/Library/Python/2.7/lib/python/site-packages/lifetimes/utils.py", line 290, in _scale_time
    return 10. / age.max()
  File "/usr/local/lib/python2.7/site-packages/numpy/core/_methods.py", line 26, in _amax
    return umr_maximum(a, axis, None, out, keepdims)
ValueError: zero-size array to reduction operation maximum which has no identity
```

CamDavidsonPilon:
Okay, I see the problem now. Thanks for taking the time to explain it 👍

PR #117, opened 2017-07-18
Port p_alive fix for Pareto_NBD from Theofilos BTYD repo.

statwonk:
I noticed two outstanding issues related to numerical optimization stability: https://github.com/CamDavidsonPilon/lifetimes/issues/19 and https://github.com/CamDavidsonPilon/lifetimes/issues/25#issuecomment-152871363. I thought to continue getting up to speed with this library, I'd try to knock out the remaining stability issue.

I wasn't familiar with using a hypergeometric series to improve stability, so this is really just a port of a fix I found in [Theofilos repository](https://github.com/theofilos/BTYD/commit/b9b89ed591858cf6cb9406c7393258f3492b7c79) (though I did start learning about hypergeometric series, very interesting!).

I'd like to add at least one or two tests, I'll ponder those tonight.

### DIFF
--- a/lifetimes/fitters/pareto_nbd_fitter.py
+++ b/lifetimes/fitters/pareto_nbd_fitter.py
@@ -136,6 +136,9 @@ class ParetoNBDFitter(BaseFitter):
         From paper:
         http://brucehardie.com/notes/009/pareto_nbd_derivations_2005-11-05.pdf
 
+        Numerical optimization authored by Ricardo Pereira,
+        https://github.com/theofilos/BTYD/blob/master/pnbd.R#L284
+
         Parameters:
             frequency: a scalar: historical frequency of customer.
             recency: a scalar: historical recency of customer.
@@ -147,9 +150,19 @@ class ParetoNBDFitter(BaseFitter):
         x, t_x = frequency, recency
         r, alpha, s, beta = self._unload_params('r', 'alpha', 's', 'beta')
 
-        A_0 = np.exp(self._log_A_0([r, alpha, s, beta], x, t_x, T))
-        return 1. / (1. + (s / (r + s + x)) *
-                     (alpha + T) ** (r + x) * (beta + T) ** s * A_0)
+        A_0 = 0
+        if alpha >= beta:
+            F1 = hyp2f1(r + s + x, s + 1, r + s + x + 1, (alpha - beta)/(alpha + t_x))
+            F2 = hyp2f1(r + s + x, s + 1, r + s + x + 1, (alpha - beta)/(alpha + T))
+            X1 = F1 * ((alpha + T)/(alpha + t_x))**(r + x) * ((beta + T)/(alpha + t_x))**s
+            X2 = F2 * ((beta + T)/(alpha + T))**s
+        else:
+            F1 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + t_x))
+            F2 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + T))
+            X1 = F1 * ((alpha + T)/(beta + t_x))**(r + x) * ((beta + T)/(beta + t_x))**s
+            X2 = F2 * ((alpha + T)/(beta + T))**(r + x)
+            return (1 + s/(r + s + x) * (X1-X2))**(-1)
+        return (1 + s/(r + s + x) * (X1-X2))**(-1)
 
     def conditional_probability_alive_matrix(self, max_frequency=None,
                                              max_recency=None):
--- a/lifetimes/fitters/pareto_nbd_fitter.py
+++ b/lifetimes/fitters/pareto_nbd_fitter.py
@@ -150,19 +150,19 @@ class ParetoNBDFitter(BaseFitter):
         x, t_x = frequency, recency
         r, alpha, s, beta = self._unload_params('r', 'alpha', 's', 'beta')
 
-        A_0 = 0
-        if alpha >= beta:
-            F1 = hyp2f1(r + s + x, s + 1, r + s + x + 1, (alpha - beta)/(alpha + t_x))
-            F2 = hyp2f1(r + s + x, s + 1, r + s + x + 1, (alpha - beta)/(alpha + T))
-            X1 = F1 * ((alpha + T)/(alpha + t_x))**(r + x) * ((beta + T)/(alpha + t_x))**s
-            X2 = F2 * ((beta + T)/(alpha + T))**s
+        maxab = max(alpha, beta)
+        absab = abs(alpha - beta)
+        if alpha < beta:
+            param2 = r + x
         else:
-            F1 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + t_x))
-            F2 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + T))
-            X1 = F1 * ((alpha + T)/(beta + t_x))**(r + x) * ((beta + T)/(beta + t_x))**s
-            X2 = F2 * ((alpha + T)/(beta + T))**(r + x)
-            return (1 + s/(r + s + x) * (X1-X2))**(-1)
-        return (1 + s/(r + s + x) * (X1-X2))**(-1)
+            param2 = s + 1.
+
+        F0 = ((alpha + T)**(r + x)) * ((beta + T)**s)
+        F1 = hyp2f1(r + s + x, param2, r + s + x + 1., absab / (maxab + t_x)) / \
+            ((maxab + t_x)**(r + s + x))
+        F2 = hyp2f1(r + s + x, param2, r + s + x + 1., absab / (maxab + t_x)) / \
+            ((maxab + T)**(r + s + x))
+        return 1./(1. + (s/(r + s + x)) * F0 * (F1 - F2))
 
     def conditional_probability_alive_matrix(self, max_frequency=None,
                                              max_recency=None):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 numpy
 scipy
 pandas>=0.19
-dill
+dill==0.2.7.1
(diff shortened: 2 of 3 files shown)


lifetimes/fitters/pareto_nbd_fitter.py
```diff
@@ -147,9 +150,19 @@ def conditional_probability_alive(self, frequency, recency, T):
         x, t_x = frequency, recency
         r, alpha, s, beta = self._unload_params('r', 'alpha', 's', 'beta')
 
-        A_0 = np.exp(self._log_A_0([r, alpha, s, beta], x, t_x, T))
-        return 1. / (1. + (s / (r + s + x)) *
-                     (alpha + T) ** (r + x) * (beta + T) ** s * A_0)
+        A_0 = 0
+        if alpha >= beta:
```
statwonk: Heh, I can see I'm not being very pythonic here in contrast with `_log_A_0`, https://github.com/CamDavidsonPilon/lifetimes/blob/227b64a35f9006c154d110a68b3268fc89ad1977/lifetimes/fitters/pareto_nbd_fitter.py#L86. I'm gonna give that a shot.

CamDavidsonPilon: what's not pythonic about it? :)

CamDavidsonPilon: Actually, the two return statements, with on inside the `else` and the other outside, is strange. I would move the second one, `return (1 + s/(r + s + x) * (X1-X2))**(-1)`, into the `if` part.

CamDavidsonPilon: hm, on second look, they look identical (unless I am missing something), so I would remove the one inside the `else`.

lifetimes/fitters/pareto_nbd_fitter.py
```diff
@@ -147,9 +150,19 @@ def conditional_probability_alive(self, frequency, recency, T):
...
+            F2 = hyp2f1(r + s + x, s + 1, r + s + x + 1, (alpha - beta)/(alpha + T))
+            X1 = F1 * ((alpha + T)/(alpha + t_x))**(r + x) * ((beta + T)/(alpha + t_x))**s
+            X2 = F2 * ((beta + T)/(alpha + T))**s
+        else:
+            F1 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + t_x))
+            F2 = hyp2f1(r + s + x, r + x, r + s + x + 1, (alpha - beta)/(beta + T))
+            X1 = F1 * ((alpha + T)/(beta + t_x))**(r + x) * ((beta + T)/(beta + t_x))**s
+            X2 = F2 * ((alpha + T)/(beta + T))**(r + x)
+            return (1 + s/(r + s + x) * (X1-X2))**(-1)
+        return (1 + s/(r + s + x) * (X1-X2))**(-1)
```
CamDavidsonPilon: instead of the exponent `-1`, I would explicitly use the division instead:
```
1/ (1 + s/(r + s + x) * (X1-X2))
```

lifetimes/fitters/pareto_nbd_fitter.py
```diff
@@ -147,9 +150,19 @@ def conditional_probability_alive(self, frequency, recency, T):
         x, t_x = frequency, recency
         r, alpha, s, beta = self._unload_params('r', 'alpha', 's', 'beta')
 
-        A_0 = np.exp(self._log_A_0([r, alpha, s, beta], x, t_x, T))
-        return 1. / (1. + (s / (r + s + x)) *
-                     (alpha + T) ** (r + x) * (beta + T) ** s * A_0)
+        A_0 = 0
```
CamDavidsonPilon: what does this do?

CamDavidsonPilon:
Looking good! Good start is that all current tests still pass.

lifetimes/fitters/pareto_nbd_fitter.py
statwonk: Heh, nothing. Accidentally cribbed it along with the fix! 😬  I wish I knew exactly how `hyp2f1` worked, the rabbit hole looks mighty deep!

statwonk: Ah, I missed this nice explicit version, [much more understandable](https://github.com/theofilos/BTYD/blob/master/pnbd.R#L6).

statwonk:
Heh, nothing's ever easy, right? The new test reveals my fix here produces slightly different estimates.

The fix produces P(alive) of `0.9994` vs `0.9979` (target). I think I might do some research to see if anyone else has attempted a fix on the R side.

Save that, I'm not sure how else to prove it's correct. 🤔

lifetimes/fitters/pareto_nbd_fitter.py
statwonk: They're slightly different, for example the third argument of `hyp2f1` (see denominator). I'm going to see if I can factor or at least understand how this numerical optimization works. I found [one additional source](http://www.brucehardie.com/notes/008/pareto_nbd_MATLAB.pdf).

statwonk:
Alright! Found this source, now the anchor test passes: http://www.brucehardie.com/notes/008/pareto_nbd_MATLAB.pdf

statwonk:
Well darn, perhaps the best we can offer here is a clear error?

In the test case we're trying to exponentiate to ~`^405`,

```
import decimal
x = Decimal(7.98565181e-03+7) # from test case
x**405
Decimal('1.726957276326162965747888453E+338')
np.float64(Decimal('2.919090862580816514432675122E+342'))
inf
```

Even `np.float64` gives up.

I could easily see use-cases where frequency might be greater than 400, so it seems that this is a fairly built-in limitation of Pareto/NBD. Perhaps this is Fader's motivation in moving to BG/NBD?

requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy
 scipy
 pandas>=0.19
-dill
\ No newline at end of file
+dill==0.2.7.1
```
CamDavidsonPilon: this needed?

lifetimes/fitters/pareto_nbd_fitter.py
```diff
@@ -147,9 +150,19 @@ def conditional_probability_alive(self, frequency, recency, T):
         x, t_x = frequency, recency
         r, alpha, s, beta = self._unload_params('r', 'alpha', 's', 'beta')
 
-        A_0 = np.exp(self._log_A_0([r, alpha, s, beta], x, t_x, T))
-        return 1. / (1. + (s / (r + s + x)) *
-                     (alpha + T) ** (r + x) * (beta + T) ** s * A_0)
+        maxab = max(alpha, beta)
```
CamDavidsonPilon: 👍 nice looking code here

requirements.txt
statwonk: No, I had two failing tests caused by the version of `dill` I had installed. I upgraded to the newest version and it was fixed, so I figured I'd pin it to known working version. I'll unpin. :) In general, I'm curious why you prefer unpinned?

tests/test_estimation.py
```diff
@@ -302,6 +302,29 @@ def test_conditional_probability_alive_is_between_0_and_1(self, cdnow_customers)
...
+            zip(['r', 'alpha', 's', 'beta'],
+            [0.5534, 10.5802, 0.6061, 11.6562]))
+        p_alive = ptf.conditional_probability_alive(26.00, 30.86, 31.00)
+        assert abs(p_alive - 0.9979) < 0.001
+
+    def test_conditional_probability_alive_overflow_error(self):
+        ptf = estimation.ParetoNBDFitter()
+        ptf.params_ = OrderedDict(
+            zip(['r', 'alpha', 's', 'beta'],
+            [10.465, 7.98565181e-03, 3.0516, 2.820]))
```
statwonk: I nabbed them from here: https://github.com/CamDavidsonPilon/lifetimes/blob/cdd307f648a778088d5bdd3fd6c0e3f236bbe8a0/tests/test_estimation.py#L233

lifetimes/fitters/pareto_nbd_fitter.py
```diff
@@ -147,9 +150,19 @@ def conditional_probability_alive(self, frequency, recency, T):
...
+        if alpha < beta:
+            param2 = r + x
+        else:
+            param2 = s + 1.
+
+        F0 = ((alpha + T)**(r + x)) * ((beta + T)**s)
+        F1 = hyp2f1(r + s + x, param2, r + s + x + 1., absab / (maxab + t_x)) / \
+            ((maxab + t_x)**(r + s + x))
+        F2 = hyp2f1(r + s + x, param2, r + s + x + 1., absab / (maxab + t_x)) / \
+            ((maxab + T)**(r + s + x))
```
statwonk: @CamDavidsonPilon the best solution I can see after today's effort is a check that protects against `(recency + frequency) > some_constant` (where `some_constant` is about 300 or so). The problem is that `(maxab + T)` is a `float` and when raised to a very large exponent floating point arithmetic fails. Some cursory investigation of `numpy` and `mpmath` didn't yield a solution.

CamDavidsonPilon:
So I added the two new tests into my local master and tested them, and both were 🍏. So maybe the function is already very robust against overflow and we don't need to change them anymore. WDYT?

statwonk:
Yeah, that could be. I should have started with the tests. :-p I'll close this PR.

lifetimes/fitters/pareto_nbd_fitter.py
achiru: posted on the other page and reposting it here for consistency:

guess this is easier to write here, instead of pull request.

I think instead of trying the computational approach, if we use the alternative and more simplified expression for P_Alive which is Eq. 34 in the reference: http://brucehardie.com/notes/009/pareto_nbd_derivations_2005-11-05.pdf, this problem will be fixed.

I tried it , and it eliminates the overflow for high values of R,F, and T (verified for range of values >300).

below is the function with expression in 34. Note that parts of this equation as indeed present in conditional expected number of purchases.

```
def conditional_probability_alive_2(model, frequency, recency, T):
        params=model.params_.values()
        r,alpha,s,beta= params
        x, t_x = frequency, recency
        likelihood = -model._negative_log_likelihood(params, x, t_x, T,0.0) 
        first_term = gammaln(r + x) - gammaln(r) + r * log(alpha) + s * \
            log(beta) - (r + x) * log(alpha + T) - s * log(beta + T) 
        return exp(first_term - likelihood)
```



Thank you very much for the package.
Aravind.

PR #120, opened 2017-07-20
Make observation_period_end default to the max transaction time

statwonk:
This PR attempts making the api a bit less surprising by setting `observation_period_end` to the max transaction time in the user-supplied transaction data.

Conversation about this change in default behavior can be found here: https://github.com/CamDavidsonPilon/lifetimes/issues/67.


lifetimes/utils.py
```diff
@@ -148,7 +148,7 @@ def find_first_transactions(transactions, customer_id_col, datetime_col, monetar
 
 
 def summary_data_from_transaction_data(transactions, customer_id_col, datetime_col, monetary_value_col=None, datetime_format=None,
-                                       observation_period_end=datetime.today(), freq='D'):
+                                       observation_period_end=None, freq='D'):
```
CamDavidsonPilon: can you add docs for this?

statwonk: Will do! Thanks for the prod!

CamDavidsonPilon:
lgtm! Merging now. Closes #67

PR #122, opened 2017-07-22
Docs

CamDavidsonPilon:
@aprotopopov let's merge this. I think the initial quickstart bits we have are good for a public release, and we can add iteratively. WDYT? If you have any outstanding contributions, please add them here soon.


README.md
```diff
@@ -1,10 +1,13 @@
 ![](http://i.imgur.com/7s3jqZM.png)
```
CamDavidsonPilon: need a link to docs here

aprotopopov:
Yeah, let's merge. No any additional commits left.

PR #126, opened 2017-08-15
Tests plots

aprotopopov:
Add tests for 4 plots:
- plot_cumulative_transactions
- plot_incremental_transactions
- plot_transaction_rate_heterogeneity
- plot_dropout_rate_heterogeneity

Two of them are slow due to not optimal [`expected_cumulative_transactions`](https://github.com/CamDavidsonPilon/lifetimes/blob/master/lifetimes/utils.py#L401) realization (I think it could be improved):

```
=============================================== slowest 5 test durations ===============================================
10.73s call     tests/test_plotting.py::TestPlotting::test_plot_incremental_transactions
10.69s call     tests/test_plotting.py::TestPlotting::test_plot_cumulative_transactions
4.36s call     tests/test_estimation.py::TestModifiedBetaGammaFitter::test_purchase_predictions_do_not_differ_much_if_looking_at_hourly_or_daily_frequencies
3.64s call     tests/test_estimation.py::TestParetoNBDFitter::test_params_out_is_close_to_Hardie_paper
3.52s call     tests/test_plotting.py::TestPlotting::test_plot_calibration_purchases_vs_holdout_purchases
```

P.S. Still don't know how to manage travis for 2 RMS tolerance, so insert constant `TOLERANCE_VALUE` and set value to 20. But it should be decreased for better tests.


tests/test_utils.py
```diff
@@ -247,13 +246,11 @@ def test_summary_statistics_are_indentical_to_hardies_paper_confirming_correct_a
     # see http://brucehardie.com/papers/rfm_clv_2005-02-16.pdf
     # RFM and CLV: Using Iso-value Curves for Customer Base Analysis
     df = pd.read_csv('lifetimes/datasets/CDNOW_sample.txt', sep='\s+', header=None, names=['_id', 'id', 'date', 'cds_bought', 'spent'])
-    if sys.version_info[0] < 3:
-        df['date'] = pd.to_datetime(df['date'].astype(unicode))
-    else:
-        df['date'] = pd.to_datetime(df['date'], format='%Y%m%d')
+    df['date'] = pd.to_datetime(df['date'], format='%Y%m%d')
```
CamDavidsonPilon: why this change?

CamDavidsonPilon:
New plots + tests look great 👍 

One comment about a python3 change, otherwise lgtm

tests/test_utils.py
aprotopopov: Because that working both for python2 and python3 without extra if/else statements. To make code cleaner. Not sure is .astype(unicode)` necessary for python2.

CamDavidsonPilon: it does seem strange...and if tests pass, then it's our own fault it we are missing something.

aprotopopov: I still don't understand why there should be errors or what wrong with that change. Tested with python2 and python3:

```python
import pandas as pd
df = pd.read_csv('lifetimes/datasets/CDNOW_sample.txt', sep='\s+', header=None, names=['_id', 'id', 'date', 'cds_bought', 'spent'])
df.head()
"""
	_id 	id 	date 	cds_bought 	spent
0 	4 	1 	19970101 	2 	29.33
1 	4 	1 	19970118 	2 	29.73
2 	4 	1 	19970802 	1 	14.96
3 	4 	1 	19971212 	2 	26.48
4 	21 	2 	19970101 	3 	63.34
"""

pd.to_datetime(df['date'], format='%Y%m%d').head()
"""
0   1997-01-01
1   1997-01-18
2   1997-08-02
3   1997-12-12
4   1997-01-01
Name: date, dtype: datetime64[ns]
"""
```

CamDavidsonPilon: I agree - what I meant was that a regression test should have been added at the same time to fix whatever broke between 2 and 3. Though clearly you have provided evidence that probably nothing is broken!

PR #129, opened 2017-08-16
Save model docs

aprotopopov:
Docs for current saving and loading model routine. Several thoughts about next:
- move `save_model` and `load_model` to function in something like `utils.py`
- add `to_json` and `to_yaml` functions to save only parameters of models
- add `save_data` to save only data

May be `save_model` could be rewritten to have an option how to save with three options:
- dill
- JSON
- YAML

### DIFF
--- a/docs/Save and load model.md
+++ b/docs/Save and load model.md
@@ -1,0 +1,35 @@
+## Save and load model
+
+When you have lots of data and training takes a lot of time option with saving and loading model could be useful. First you need fit model, then save it and load.
+
+### Fit model
+
+```python
+from lifetimes import BetaGeoFitter
+from lifetimes.datasets import load_cdnow_summary
+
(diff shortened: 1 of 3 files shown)


docs/Save and load model.md
```diff
@@ -0,0 +1,35 @@
+## Save and load model
```
CamDavidsonPilon: Saving and loading models

docs/Save and load model.md
```diff
@@ -0,0 +1,35 @@
...
+
+When you have lots of data and training takes a lot of time option with saving and loading model could be useful. First you need fit model, then save it and load.
```
CamDavidsonPilon: "you need to fit the model"

aprotopopov:
Yeah, silly mistakes...

PR #130, opened 2017-08-17
Model routines

aprotopopov:
Add option to save or not `generate_new_data` method. Without saving `data` and `generate_new_data` final model object is small. Could be usefull if you need only parameters of model.


aprotopopov:
P.S. what should be saved with `to_json` and `to_yaml` method? I think only parameters without heavy objects. Something like that (may be with corrections):

```python
import json
from lifetimes import BetaGeoFitter
from lifetimes.datasets import load_cdnow_summary

data = load_cdnow_summary(index_col=[0])
bgf = BetaGeoFitter()
bgf.fit(data['frequency'], data['recency'], data['T'])

print(json.dumps(bgf, default=lambda o: o.__dict__, sort_keys=True, indent=4))
"""
{
    "_negative_log_likelihood_": 2.6505481706750498,
    "_scale": 0.2573340195573855,
    "data": {
        "_data": {},
        "_item_cache": {},
        "is_copy": null
    },
    "generate_new_data": {},
    "params_": {
        "a": 0.7928754636277693,
        "alpha": 4.413484046244373,
        "b": 2.4257092100849764,
        "r": 0.24259201144750225
    },
    "penalizer_coef": 0.0,
    "predict": {}
}
"""
```

stochastic1:
Would you need to maintain a shape parameter when there's no data committed for __repr__?

lifetimes/fitters/base_fitter.py
```diff
@@ -0,0 +1,61 @@
...
+        save_generate_data_method: bool, optional
+            Whether to save generate_new_data method (if it exists) from
+            fitter.generate_new_data to pickle object.
+        values_to_save: list, optional
+            Placeholders for original attributes for saving object. If None
+            will be extended to attr_list length like [None] * len(attr_list)
+
+        """
+        attr_list = ['data' * (not save_data),
+                     'generate_new_data' * (not save_generate_data_method)]
```
CamDavidsonPilon: hm neat way of doing this

lifetimes/utils.py
```diff
@@ -466,3 +468,36 @@ def expected_cumulative_transactions(model, transactions, datetime_col, customer
...
+        Placeholders for original attributes for saving object. If None will be
+        extended to attr_list length like [None] * len(attr_list)
+    """
+    if values_to_save is None:
+        values_to_save = [None] * len(attr_list)
+
+    saved_attr_dict = {}
+    for attr, val_save in zip(attr_list, values_to_save):
+        if attr in obj.__dict__:
+            item = obj.__dict__.pop(attr)
```
CamDavidsonPilon: is this going to mutate the original `obj`?

CamDavidsonPilon: I thought it would, but the tests suggest otherwise...

lifetimes/fitters/base_fitter.py
```diff
@@ -0,0 +1,61 @@
...
+    """Base class for fitters."""
+
+    def __repr__(self):
+        """Representation of fitter."""
+        classname = self.__class__.__name__
+        try:
+            param_str = ", ".join("%s: %.2f" % (param, value) for param, value
+                                  in sorted(self.params_.items()))
+            return "<lifetimes.%s: fitted with %d subjects, %s>" % (
+                classname, self.data.shape[0], param_str)
```
CamDavidsonPilon: @stochastic1 brings up a good point - does repr still work if we don't save the data with it? Perhaps we need to protect against this case too.

aprotopopov: No, repr changed to just fitter name. I'll changed it soon.

lifetimes/utils.py
aprotopopov: But then attr object returns back to `obj` with:
```
for attr, item in saved_attr_dict.items():
    setattr(obj, attr, item)
```

aprotopopov:
If there are no any additional suggestions/comments I think PR could be merged.

PR #131, opened 2017-08-18
travis simplified, new python version 3.6-dev added

aprotopopov:
With python 3.7-dev there are [errors](https://travis-ci.org/CamDavidsonPilon/lifetimes/jobs/265838722), so excluded it. 
Also explicitly add dist trusty for .travis.yml because they are [planing](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming) to change default distro to trusty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,15 @@
 language: python
+dist: trusty
 python:
    - "2.7"
    - "3.4"
    - "3.5"
    - "3.6"
+   - "3.6-dev" # 3.6 development branch
 before_install:
   - sudo apt-get update
-# See http://conda.pydata.org/docs/travis.html
 install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,35 +1,15 @@
...
-  - conda update conda
-  # Useful for debugging any issues with conda
-  - conda info -a
-
-  # Should match requirements.txt
-  - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION pip numpy scipy pandas matplotlib nose pytest dill
-  - source activate test-environment
-  - python setup.py install
-  - pip install coveralls
-  - pip install pytest-cov pytest-mpl==0.5
```
CamDavidsonPilon: 💯 💯 

Such an obvious change

Issue #144, opened 2017-11-03
conditional_expected_number_of_purchases_up_to_time shows weird values for ParetoNBDFitter


mwrabel:
It seems that conditional_expected_number_of_purchases_up_to_time method works weird for ParetoNBDFitter. Users with positive frequency and recency experience near zero estimate for predicted purchases, while those with frequency and recency =0 have higher values of predicted purchases. Not sure if it's a bug though, are you able to confirm?

You can easily replicate this with code below:
```python
from lifetimes.datasets import load_cdnow_summary
data = load_cdnow_summary(index_col=[0])

print(data.head())

from lifetimes import BetaGeoFitter
from lifetimes import ParetoNBDFitter

bgf = BetaGeoFitter(penalizer_coef=0.0) # 0.001 to 0.1 are effective (L2 regularization)
bgf.fit(data['frequency'], data['recency'], data['T'])
print(bgf)
pnbd = ParetoNBDFitter(penalizer_coef=0.0) # 0.001 to 0.1 are effective (L2 regularization)
pnbd.fit(data['frequency'], data['recency'], data['T'])
print(pnbd)

t = 30
data['pred_purchases_1m_bgf'] = bgf.conditional_expected_number_of_purchases_up_to_time(t, data['frequency'], data['recency'], data['T'])
data['pred_purchases_1m_pnbd'] = pnbd.conditional_expected_number_of_purchases_up_to_time(t, data['frequency'], data['recency'], data['T'])
print data.head(20)
```

![result](https://user-images.githubusercontent.com/17828920/32366921-851fb9e0-c080-11e7-9130-276fd74b0a3f.JPG)

rhydomako:
I think I agree that there is a bug in `ParetoNBDFitter. conditional_expected_number_of_purchases_up_to_time`.

Starting with the above fitted model, take a look at what happens when I look at just a couple of rows:
```
t = 30
i = 3

pnbd.conditional_expected_number_of_purchases_up_to_time(
    t, 
    data['frequency'].head(i), 
    data['recency'].head(i),
    data['T'].head(i)
)
```

```
ID
1    0.010421
2    0.201782
3    6.422322
dtype: float64
```

Now, if I append a couple more rows to the conditional expectation calculation:
```
t = 30
i = 5

pnbd.conditional_expected_number_of_purchases_up_to_time(
    t, 
    data['frequency'].head(i), 
    data['recency'].head(i),
    data['T'].head(i)
)
```

```
[...]

PR #145, opened 2017-11-04
Decouple ParetoNBD log-likelihood calculation from optimization step

rhydomako:
As discussed in #144 `ParetoNBDFitter.conditional_expected_number_of_purchases_up_to_time` uses the conditional log-likelihood to perform its calculation. However, the `_negative_log_likelihood` method actually marginalizes across all of the rows as part of the full negative log-likelihood calculation for the parameter optimization step.

To address this I've made a new function `_conditional_log_likelihood` which makes the intermediate calculation that both `_negative_log_likelihood` and `conditional_expected_number_of_purchases_up_to_time` require.

### DIFF
--- a/lifetimes/fitters/pareto_nbd_fitter.py
+++ b/lifetimes/fitters/pareto_nbd_fitter.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 from collections import OrderedDict
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
ah, annoying image failures. Yea this looks great, thanks!

PR #153, opened 2017-12-18
WIP - Add a weight kwarg to fit

AustinRochford:
When there are many fewer (`recency`, `frequency`, `T`) triplets than there are customers, weighting the likelihood terms instead of duplicating them can drastically speed up calculation (I've seen a 20x speedup on real-world ecomm data).

This is just a WIP to see how hard this is and gauge interest in this as a contribution.  I'll add it to all models and clean up the interface and implementation (when I have time) if there is interest.

### DIFF
--- a/lifetimes/fitters/beta_geo_fitter.py
+++ b/lifetimes/fitters/beta_geo_fitter.py
@@ -54,7 +54,7 @@ class BetaGeoFitter(BaseFitter):
 
     def fit(self, frequency, recency, T, iterative_fitting=1,
             initial_params=None, verbose=False, tol=1e-4, index=None,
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
hey @AustinRochford, 

I'm very interested in this PR. I've been tinkering with a similar idea in my head for another project, so this is very relevant for me. Please continue! Can you show an example of the preprocessing of the raw customer summary data?

AustinRochford:
Awesome.

I'll see if I have some time to push this to a mergeable state over the holidays.  Will include a preprocessing example for sure.

CamDavidsonPilon:
This was added in a previous release - thanks for the inspiration @AustinRochford!

Issue #169, opened 2018-03-08
BetaGeoFitter.conditional_probability_alive returns 0D-array, not float


mszep:
The docstring mentions that the return type of `BetaGeoFitter.conditional_probability_alive` is `float`, but when passing values from DataFrames, the result is actually a 0-dimensional `ndarray`:

```
In [101]: bgf = BetaGeoFitter()

In [102]: df = load_cdnow_summary()

In [103]: bgf.fit(df['frequency'],df['recency'],df['T'])
Out[103]: <lifetimes.BetaGeoFitter: fitted with 2357 subjects, a: 0.79, alpha: 4.41, b: 2.43, r: 0.24>

In [104]: bgf.conditional_probability_alive(df.iloc[0]['frequency'],df.iloc[0]['recency'
     ...: ],df.iloc[0]['T'])
Out[104]: array(0.7266095443377226)

In [105]: 
```

The same method in other fitters seem to all have forms that would indeed return `float`s, but I haven't tested that explicitly.

I can make a PR where either the docstring is altered, or the return value is wrapped with `float()`; any thoughts on which would be preferable?

CamDavidsonPilon:
Hah, good catch. I'm guessing it's because `df.iloc[0]['frequency']` and co. are not floats (or are they...). If you'd like to PR something, I think changing the return value is 😗👌

mszep:
They are floats! I had the same thought but passing literal floats doesn't change the result type. The cause is returning the result of a call to `np.where`:
```
In [13]: np.where(1==1, 2, 3)
Out[13]: array(2)
```
Anyway, PR incoming.

PR #170, opened 2018-03-09
Update beta_geo_fitter.py

mszep:
Addresses #169


CamDavidsonPilon:
We can't force a `float` here since the output could be arrays - see the failing tests in Travis CI for an example. 

We may have to check the input arrays / output array, and transform based on that.

mszep:
Looking at this again, I'm thinking it might make more sense to just change the docstring to say inputs are array-like, and output is an array. I like the flexibility of being able to pass inputs for multiple users at once.

However, then it might make sense to alter the docstrings for the other fitters as well. From an initial scan, it seems they should be able to broadcast okay, so handling array-like input shouldn't be a problem. Thoughts?

PR #176, opened 2018-03-27
Fix history alive

DaniGate:
Computation of RFM vector in calculate_alive_path function was not consistent with that in summary_from_transaction_data function:
- Counting periods with purchases instead of actual purchases in frequency column
- Age column did not reach the requested horizon t, it fell short of 1 period. We extend by 1 period the number of rows in the generated dataframe.


CamDavidsonPilon:
Shoot I totally missed this PR, apologies. 

This PR looks great - thanks for the edits! 

The failing tests are because the image has now changed - that's fine, I'll fix later. Merging!

PR #188, opened 2018-05-02
typo

EamonKeane:


CamDavidsonPilon:
👍 thanks!

EamonKeane:
No problem... just tested the library on some real world and it worked magically! Thanks for the effort you've put in. 

I've got it working in JupyterLab now, any tips/article on how to move to production?

CamDavidsonPilon:
I actually suggest updating daily, as new data comes in that could put "inactive" customers back to "active" again. If performance is an issue, I also suggest subsampling if you have over 50k customers, that's about where std. errors become quite small.

EamonKeane:
Grand, thanks I'll do that. I do have some with 200k customers but it's runs ok. Resources aren't an issue - on GKE with pre-emption the cost of running the computation is 80% off without disturbing core workload.s

PR #200, opened 2018-07-01
Removing reference plots

rhydomako:
Since the plotting reference tests have become a bit tedious to maintain, this PR goes a different direction by removing the reference comparisons. Instead, the tests will check if the figure data is making it into the matplotlib internals.

This follows the approach outlined in this post: https://stackoverflow.com/questions/27948126/how-can-i-write-unit-tests-against-code-that-uses-matplotlib/27950953#27950953. That is, we trust that matplotlib will take care of actually plotting the figure, and just test that the data/titles/legend/etc make it into the pyplot objects. This should make the test more robust to stylistic changes that happen from time to time in matplotlib.

Note that these tests aren't comprehensive. I don't think I've added checks for every element, and in some cases I only check a row or column of a much larger matrix plot. Also, since some plots are probabilistic, most data comparisons involve somewhat arbitrary tolerances (but that was also true with the reference plot scheme).


CamDavidsonPilon:
👍 yea maintaining them is too painful - maybe we will return to it in the future!

PR #203, opened 2018-08-22
feat: add new function to generate bg model transactional data

keyanyang:
Add new function to generate bg/nbd model transactional data. Currently the function in `generate_data.py` can only be used to generate summary data. There is no way to generate artificial data according to the BG/NBD model showing the details of transactions. This pull request includes the function to generate such transactional data.

Sample result:
```
  customer_id                       date
0           0 2018-12-31 00:00:00.000000
1           0 2019-01-19 05:12:29.398539
2           0 2019-02-06 10:24:58.797077
3           1 2018-12-31 00:00:00.000000
4           2 2018-12-31 00:00:00.000000
```

### DIFF
--- a/lifetimes/generate_data.py
+++ b/lifetimes/generate_data.py
@@ -56,7 +56,7 @@ def beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
             alive = np.random.random() > p
 
         times = np.array(times).cumsum()
-        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
+        df.iloc[i] = np.unique(np.array(times).astype(int)).shape[0], np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
 
     return df.set_index('customer_id')
 
@@ -179,7 +179,7 @@ def pareto_nbd_model(T, r, alpha, s, beta, size=1):
             next_purchase_in = stats.expon.rvs(scale=1. / l)
 
         times = np.array(times).cumsum()
-        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, mu, time_of_death > T[i], i
+        df.iloc[i] = np.unique(np.array(times).astype(int)).shape[0], np.max(times if times.shape[0] > 0 else 0), T[i], l, mu, time_of_death > T[i], i
 
     return df.set_index('customer_id')
 
@@ -239,7 +239,7 @@ def modified_beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
             alive = np.random.random() > p
 
         times = np.array(times).cumsum()
-        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
+        df.iloc[i] = np.unique(np.array(times).astype(int)).shape[0], np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
 
     return df.set_index('customer_id')
 
--- a/lifetimes/generate_data.py
+++ b/lifetimes/generate_data.py
@@ -61,6 +61,73 @@ def beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
     return df.set_index('customer_id')
 
 
+def beta_geometric_nbd_model_transactional_data(T, r, alpha, a, b, observation_period_end='2019-1-1',
+                                                freq='D', size=1):
+    """
+    Generate artificial transactional data according to the BG/NBD model.
+
+    See [1] for model details
+
+    Parameters
+    ----------
+    T: int, float or array_like
+        The length of time observing new customers.
+    r, alpha, a, b: float
+        Parameters in the model. See [1]_
+    observation_period_end: date_like
+        The date observation ends
+    freq: string, optional
+        Default 'D' for days, 'W' for weeks, 'h' for hours
+    size: int, optional
+        The number of customers to generate
+
+    Returns
+    -------
+    DataFrame
+        The following columns:
+        'customer_id', 'date'
+
+    References
+    ----------
+    .. [1]: '"Counting Your Customers" the Easy Way: An Alternative to the Pareto/NBD Model'
+       (http://brucehardie.com/papers/bgnbd_2004-04-20.pdf)
+
+    """
+    observation_period_end = pd.to_datetime(observation_period_end)
+
+    if type(T) in [float, int]:
+        start_date = [observation_period_end - pd.Timedelta(T - 1, unit=freq)] * size
+        T = T * np.ones(size)
+    else:
+        start_date = [observation_period_end - pd.Timedelta(T[i] - 1, unit=freq) for i in range(size)]
+        T = np.asarray(T)
+
+    probability_of_post_purchase_death = stats.beta.rvs(a, b, size=size)
+    lambda_ = stats.gamma.rvs(r, scale=1. / alpha, size=size)
+
+    columns = ['customer_id', 'date']
+    df = pd.DataFrame(columns=columns)
+
+    for i in range(size):
+        s = start_date[i]
+        p = probability_of_post_purchase_death[i]
+        l = lambda_[i]
+        age = T[i]
+
+        purchases = [[i, s - pd.Timedelta(1, unit=freq)]]
+        next_purchase_in = stats.expon.rvs(scale=1. / l)
+        alive = True
+
+        while next_purchase_in < age and alive:
+            purchases.append([i, s + pd.Timedelta(next_purchase_in, unit=freq)])
+            next_purchase_in += stats.expon.rvs(scale=1. / l)
+            alive = np.random.random() > p
+
+        df = df.append(pd.DataFrame(purchases, columns=columns))
+
+    return df.reset_index(drop=True)
+
+
 def pareto_nbd_model(T, r, alpha, s, beta, size=1):
     """
     Generate artificial data according to the Pareto/NBD model.
--- a/tests/test_generate_data.py
+++ b/tests/test_generate_data.py
@@ -8,7 +8,8 @@ import scipy.stats as stats
 
 import lifetimes.estimation as estimation
 from lifetimes.generate_data import beta_geometric_nbd_model, pareto_nbd_model, modified_beta_geometric_nbd_model, \
-    beta_geometric_beta_binom_model
+    beta_geometric_beta_binom_model, beta_geometric_nbd_model_transactional_data
+from lifetimes.utils import summary_data_from_transaction_data
 
 
 def setup_module(module):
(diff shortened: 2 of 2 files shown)


keyanyang:
Test the new function using the existing one.
In both function, `np.random.seed(188898)` is added to make sure consistency. 
```
# add seed
def beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
    if type(T) in [float, int]:
        T = T * np.ones(size)
    else:
        T = np.asarray(T)

    np.random.seed(188898)
    probability_of_post_purchase_death = stats.beta.rvs(a, b, size=size)
    np.random.seed(188898)
    lambda_ = stats.gamma.rvs(r, scale=1. / alpha, size=size)

    columns = ['frequency', 'recency', 'T', 'lambda', 'p', 'alive', 'customer_id']
    df = pd.DataFrame(np.zeros((size, len(columns))), columns=columns)

    for i in range(size):
        p = probability_of_post_purchase_death[i]
        l = lambda_[i]

        # hacky until I can find something better
        times = []
        np.random.seed(188898)
        next_purchase_in = stats.expon.rvs(scale=1. / l)
        alive = True
        while (np.sum(times) + next_purchase_in < T[i]) and alive:
            times.append(next_purchase_in)
            np.random.seed(188898)
            next_purchase_in = stats.expon.rvs(scale=1. / l)
            np.random.seed(188898)
            alive = np.random.random() > p

        times = np.array(times).cumsum()
        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i

    return df.set_index('customer_id')


def beta_geometric_nbd_model_transactional_data(T, r, alpha, a, b, start_date='2019-1-1', size=1):
    if type(T) in [float, int]:
        T = T * np.ones(size)
    else:
        T = np.asarray(T)

    start_date = pd.to_datetime(start_date)
    np.random.seed(188898)
    probability_of_post_purchase_death = stats.beta.rvs(a, b, size=size)
    np.random.seed(188898)
    lambda_ = stats.gamma.rvs(r, scale=1. / alpha, size=size)

    columns = ['customer_id', 'date']
    df = pd.DataFrame(columns=columns)

    for i in range(size):
        p = probability_of_post_purchase_death[i]
        l = lambda_[i]
        age = T[i]

        purchases = [[i, start_date - pd.DateOffset(days=1)]]  # where I change
        np.random.seed(188898)
        next_purchase_in = stats.expon.rvs(scale=1. / l)
        alive = True

        while next_purchase_in < age and alive:
            purchases.append([i, start_date + pd.DateOffset(days=next_purchase_in)])
            np.random.seed(188898)
            next_purchase_in += stats.expon.rvs(scale=1. / l)
            np.random.seed(188898)
            alive = np.random.random() > p

        df = df.append(pd.DataFrame(purchases, columns=columns))

    return df.reset_index(drop=True)
```

keyanyang:
Build test:
```
bbgb_params = OrderedDict([('T', 100), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),
                           ('b', 2.43), ('size', 100)])
summary_original = beta_geometric_nbd_model(**bbgb_params)[['frequency', 'recency', 'T']]
summary_test = summary_data_from_transaction_data(beta_geometric_nbd_model_transactional_data(**bbgb_params),
'customer_id', 'date')

summary_original['recency'] = summary_original['recency'].apply(np.ceil) # summary_data_from_transaction_data makes regency floor
summary_original = summary_original.reset_index(drop=True)
summary_test = summary_test.reset_index(drop=True)
print("\n")
print(summary_original.head())
print(summary_test.head())
print("Difference between two Dataframes:", assert_frame_equal(summary_original, summary_original))
print("Dataframes are equal:",summary_original.equals(summary_original))
```
The result:
```
   frequency  recency      T
0        5.0     92.0  100.0
1        0.0      0.0  100.0
2        0.0      0.0  100.0
3        3.0     77.0  100.0
4       10.0     92.0  100.0
   frequency  recency      T
0        5.0     92.0  100.0
1        0.0      0.0  100.0
2        0.0      0.0  100.0
3        3.0     77.0  100.0
4       10.0     92.0  100.0
Difference between two Dataframes: None
Dataframes are equal: True
```

In conclusion, this function can output **the same result after summarized** as the existing function `beta_geometric_nbd_model`. And this function can provide a way to show the **details of transactions**. This could be used for **simulation** and exploration.

lifetimes/generate_data.py
```diff
@@ -61,6 +61,66 @@ def beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
     return df.set_index('customer_id')
 
 
+def beta_geometric_nbd_model_transactional_data(T, r, alpha, a, b, start_date='2019-1-1', size=1):
```
CamDavidsonPilon: should we be opinionated about the start date?

CamDavidsonPilon: In fact, is it possible to generalize this to any time units (not just days)?

CamDavidsonPilon:
^ I love these reproducible tests - are you able to add something similar to lifetimes' test suite? Ex: https://github.com/CamDavidsonPilon/lifetimes/blob/master/tests/test_generate_data.py

lifetimes/generate_data.py
keyanyang: Yes, any valid date format can be used here. For example `'2019-1-1 12:30:21'`.

keyanyang:
I just added the test. Thank you for the comments and review Cameron!

keyanyang:
Comment for commit #203

In `generate_data.py`, currently we use `len(times)` to generate `frequency`. This will cause incorrect data.

The definition of [frequency](https://lifetimes.readthedocs.io/en/latest/Quickstart.html#the-shape-of-your-data)
> it’s the count of days the customer had a purchase on.

e.g.
If the list of times is like [2, 0.1, 0.5 0.3], the frequency should be 2 because three of these transactions happen in the same day. But existing functions will take frequency as length of times, which is 4. 

This commit is used to solve this issue.

lifetimes/generate_data.py
```diff
@@ -56,11 +56,71 @@ def beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
...
+
+    columns = ['customer_id', 'date']
+    df = pd.DataFrame(columns=columns)
+
+    for i in range(size):
+        p = probability_of_post_purchase_death[i]
+        l = lambda_[i]
+        age = T[i]
+
+        purchases = [[i, start_date - pd.DateOffset(days=1)]]
```
CamDavidsonPilon: mm but what if the user is using a different frequency, like hours or weeks?

tests/test_generate_data.py
```diff
@@ -65,3 +66,24 @@ def test_params_same_from_sim_data(self):
...
+        summary_original = beta_geometric_nbd_model(**bg_params)[['frequency', 'recency', 'T']]
+        np.random.seed(188898)
+        summary_test = summary_data_from_transaction_data(beta_geometric_nbd_model_transactional_data(**bg_params),
+        'customer_id', 'date')
+
+        summary_original['recency'] = summary_original['recency'].apply(np.ceil)  # summary_data_from_transaction_data makes regency floor
+        summary_original = summary_original.reset_index(drop=True)
+        summary_test = summary_test.reset_index(drop=True)
+        # assert
+        summary_original.equals(summary_test)
```
CamDavidsonPilon: `assert summary_original.equals(summary_test)`

lifetimes/generate_data.py
keyanyang: Thanks. I haven't considered this before. I've added `freq` argument in this function that could be used for different frequency and I've tested the use of it.

tests/test_generate_data.py
keyanyang: I've added `assert` in each test. Also I've added more tests for different type of `T` and different `freq`. Thank you!

lifetimes/generate_data.py
```diff
@@ -112,7 +179,7 @@ def pareto_nbd_model(T, r, alpha, s, beta, size=1):
             next_purchase_in = stats.expon.rvs(scale=1. / l)
 
         times = np.array(times).cumsum()
-        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, mu, time_of_death > T[i], i
+        df.iloc[i] = len(set(int(t) for t in times)), np.max(times if times.shape[0] > 0 else 0), T[i], l, mu, time_of_death > T[i], i
```
CamDavidsonPilon: This is using pure Python, which is going to be slower. What about something like:
```python
np.unique(np.array(times).astype(int)).shape[0]
```
that's about 4 times faster

lifetimes/generate_data.py
```diff
@@ -172,7 +239,7 @@ def modified_beta_geometric_nbd_model(T, r, alpha, a, b, size=1):
             alive = np.random.random() > p
 
         times = np.array(times).cumsum()
-        df.iloc[i] = len(times), np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
+        df.iloc[i] = len(set(int(t) for t in times)), np.max(times if times.shape[0] > 0 else 0), T[i], l, p, alive, i
```
CamDavidsonPilon: ditto

tests/test_generate_data.py
```diff
@@ -65,3 +66,79 @@ def test_params_same_from_sim_data(self):
                  grouped_data['n_custs'])
         assert ((np.array(list(bbgb_params.values())) - np.array(
             bbtf._unload_params('alpha', 'beta', 'gamma', 'delta'))) < 0.1).all()
+
+
+bg_params = OrderedDict([('T', 100), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),
```
CamDavidsonPilon: can you turn these into pytest fixtures? Here's a quick explanation of them: https://docs.pytest.org/en/latest/fixture.html and I use them in other areas like https://github.com/CamDavidsonPilon/lifetimes/blob/47c911a705242bbbed258372c2f06e6afcbfd428/tests/test_utils.py#L12-L59

tests/test_generate_data.py
```diff
@@ -65,3 +66,79 @@ def test_params_same_from_sim_data(self):
...
+
+
+class TestTransactionalDataGeneration():
+
+
+    # test using T as int
+    def test_beta_geometric_nbd_model_transactional_data_int_T(self):
+        np.random.seed(188898)
+        summary_original = beta_geometric_nbd_model(**bg_params)[['frequency', 'recency', 'T']]
+        np.random.seed(188898)
```
CamDavidsonPilon: why twice?

tests/test_generate_data.py
```diff
@@ -65,3 +66,79 @@ def test_params_same_from_sim_data(self):
...
+                                                          customer_id_col='customer_id', datetime_col='date',
+                                                          observation_period_end=observation_period_end, 
+                                                          freq='h')
+        summary_original['recency'] = summary_original['recency'].apply(np.ceil) # summary_data_from_transaction_data makes regency floor
+        summary_original = summary_original.reset_index(drop=True)
+        summary_test = summary_test.reset_index(drop=True)
+        assert summary_original.equals(summary_test)
+
+
+    # test using freq as week
```
CamDavidsonPilon: the function name gives a good idea of what the test is doing, you can remove these comments

tests/test_generate_data.py
keyanyang: Because using it twice to make sure both functions after these two random seeds can generate results with the same seed. I find random seed can only be applied to the next function.

tests/test_generate_data.py
```diff
@@ -65,3 +66,88 @@ def test_params_same_from_sim_data(self):
                  grouped_data['n_custs'])
         assert ((np.array(list(bbgb_params.values())) - np.array(
             bbtf._unload_params('alpha', 'beta', 'gamma', 'delta'))) < 0.1).all()
+
+
+bg_params = OrderedDict([('T', 100), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),
+                         ('b', 2.43), ('size', 800)])
+bg_params_list = OrderedDict([('T', [400, 200, 5, 103, 198, 401]), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),
+                              ('b', 2.43), ('size', 6)])
+bg_params_with_freq = OrderedDict([('T', 100), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),
```
CamDavidsonPilon: even these can be py.test fixtures:
```
@pytest.fixture
def bg_params_with_freq():
    return  OrderedDict([('T', 100), ('r', 0.24), ('alpha', 4.41),  ('a', 0.79),('b', 2.43), ('size', 800), ('freq', 'h')])

```

CamDavidsonPilon:
@keyanyang thanks for the contribution!

keyanyang:
Thank you so much! @CamDavidsonPilon

PR #128, opened 2017-08-16
Fix runtime warning of log

harrisong:
Fixes the scenario when an invalid value is put in log

`/usr/local/lib/python3.5/site-packages/lifetimes/estimation.py:645: RuntimeWarning: invalid value encountered in log
  A_4 = log(a) - log(b + freq - 1) - (r + freq) * log(rec + alpha)`

PR #206, opened 2018-09-15
Fix/invalid log conditional expectation

harrisong:
This PR removes `exp` after applying `log` (revert those lines to the [previous commit](https://github.com/CamDavidsonPilon/lifetimes/pull/113/files/784f55c6873f530b20744d2093433c28868d30f4#diff-3810f074613cf8b640b39ed950d552f5L680)), as mathematically they are the same.

However, production environment tells us that there are cases when the fitted `r`, `alpha`, `a`, `b` can produce a negative `hyp_term`. As a result, passing that into the `log` function will return invalid results.


harrisong:
@AdriMarteau will be great to know if there were any reasons that the patch was created! Any help is appreciated!

tests/test_estimation.py
```diff
@@ -384,6 +384,17 @@ def test_conditional_expectation_returns_same_value_as_Hardie_excel_sheet(self,
         actual = bfg.conditional_expected_number_of_purchases_up_to_time(t, x, t_x, T)
         assert abs(expected - actual) < 0.001
 
+    def test_conditional_expectation_with_invalid_fitted_params(self, cdnow_customers):
+        bfg = estimation.BetaGeoFitter()
+        bfg.params_ = OrderedDict([('r', 0.5458741247391189), ('alpha', 13.409316394557274), ('a', 0.0009994943799344323), ('b', 0.03899022143378801)])
```
CamDavidsonPilon: which part is invalid, to be clear?

CamDavidsonPilon:
Thanks for the contribution @harrisong - I was just about to pick up your other PR #128 :) 
(which I'll still do now)

tests/test_estimation.py
harrisong: Looks like the naming is not clear enough. This combination of params can cause a negative `hyp2f1_term`

tests/test_estimation.py
```diff
@@ -384,6 +383,28 @@ def test_conditional_expectation_returns_same_value_as_Hardie_excel_sheet(self,
...
+        t = 180
+        x = 0
+        t_x = 0
+        T = 5
+        expected = 5.212
+        actual = bfg.conditional_expected_number_of_purchases_up_to_time(t, x, t_x, T)
+        assert abs(expected - actual) < 0.001
+
+    def test_conditional_expectation_overflow_error_with_high_frequency(self, cdnow_customers):
```
CamDavidsonPilon: this test is failing in CI: https://travis-ci.org/CamDavidsonPilon/lifetimes/jobs/429049394#L705

does it pass locally?

harrisong: Yes, @CamDavidsonPilon, thanks for notifying! So essentially I realised this patch will not pass when the frequency has some high values as the`hyp2f1_term` becomes `inf`. In the original patch, the `log` trick does solve the issue, but won't accept a negative `hyp2f1_term`. So now I am stuck in this predicament. 

Any ideas how we can solve both of these problems?

harrisong: Looks like `mpmath.hyp2f1` fixes the overflow issue encountered when using `scipy.specia.hyp2f1`... Interesting....

CamDavidsonPilon (commented):
one idea could be to keep the original code, as it handles your new edge case well, but also include something like:
```python
hyp_term = hyp2f1(_a, _b, _c, _z)
hyp_term_min = hyp_term.min()
hyp_term -= hyp_term_min #  (all values are atleast 0 - may need to add a small epsilon too)
...
...

second_term = (1 - (hyp_term_min + exp(ln_hyp_term))*exp((r + x) *
                            np.log((alpha + T) / (alpha + t + T))))
# I may have messed the above brackets up a bit 
```

This also has the effect of making *all* computations a bit more stable (since values are a bit closer to 0, even in the "expected case).

Thoughts?

harrisong:
That seems to be a reasonable approach, and seems to pass the `test_conditional_expectation_with_negative_hyp2f1_term` case well.

I am wondering how it can work with the `ln_hyp_term_alt` term too.

psygo:
Any progress with this *PR*?

PR #236, opened 2018-12-11
Add Q-Constraint to Avoid Negative CLTV Values

vruvora:
Fixes negative values that come out of the CLV by penalizing q < 1.

### DIFF
--- a/lifetimes/fitters/gamma_gamma_fitter.py
+++ b/lifetimes/fitters/gamma_gamma_fitter.py
@@ -51,10 +51,14 @@ class GammaGammaFitter(BaseFitter):
 
     @staticmethod
     def _negative_log_likelihood(params, frequency, avg_monetary_value,
-                                 penalizer_coef=0):
-        if any(i < 0 for i in params):
-            return np.inf
-
+                                 penalizer_coef=0, q_constraint=True):
+        if q_constraint:
+            if any(i < 0 for i in params) or params[1] < 1:
+                return np.inf
+        else:
+            if any(i < 0 for i in params):
+                return np.inf
+        
         p, q, v = params
 
         x = frequency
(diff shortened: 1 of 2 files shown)


lifetimes/fitters/gamma_gamma_fitter.py
```diff
@@ -109,7 +113,7 @@ def conditional_expected_average_profit(self, frequency=None,
 
     def fit(self, frequency, monetary_value, iterative_fitting=4,
             initial_params=None, verbose=False, tol=1e-4, index=None,
-            fit_method='Nelder-Mead', maxiter=2000, **kwargs):
+            fit_method='Nelder-Mead', maxiter=2000, q_constraint=True, **kwargs):
```
CamDavidsonPilon: can you add a doc string to describe this parameter?

lifetimes/fitters/gamma_gamma_fitter.py
```diff
@@ -51,10 +51,14 @@ def __init__(self, penalizer_coef=0.0):
 
     @staticmethod
     def _negative_log_likelihood(params, frequency, avg_monetary_value,
-                                 penalizer_coef=0):
-        if any(i < 0 for i in params):
-            return np.inf
-
+                                 penalizer_coef=0, q_constraint=True):
+        if q_constraint:
```
CamDavidsonPilon: I think the checks of q-constraint and `any(i < 0 for i in params)` are independent, and can be checked seperately. Ex:
```python
if any(i < 0 for i in params):
    return np.inf
if q_constraint and params[1] < 1:
    return np.inf
```

CamDavidsonPilon:
Possible to add a test too?

CamDavidsonPilon:
I think existing is fine. I see two tests: one test that confirms `q_constraint` can be set to true in the `fit` and nothing breaks, and a second test against `_negative_log_likelihood` that gives `inf` when the q param is less than 1 and `q_contraint` is True. Make sense?

vruvora:
@CamDavidsonPilon So the default value is `True` and checks seem to pass, so I should probably set the default value to be `False` and have a specific case for `True`. Should not this default `True`? 

The other one makes sense.

CamDavidsonPilon:
Default value I feel should be False, I think.

vruvora:
@CamDavidsonPilon Updated.

CamDavidsonPilon:
😍 perfect, thanks for the contribution @vruvora

PR #237, opened 2018-12-13
Speedup of summary_data_from_transaction_data

MichaelSchreier:
Replaced pd.period instances with timestamps for the actual calculations / aggregations which yields a speedup of several factors x10 on large datasets. For a dummy dataset with 100000 entries execution time drops from ~7 seconds to 0.13 seconds on my machine.

.to_period is still applied to truncate dates (more consistently in fact) which means results should remain largely unaffected.

Note that starting from pandas 0.24.0 some or all of the operations involving pd.period will also be [considerably faster](https://github.com/pandas-dev/pandas/issues/18053), however, the calculations via timestamps (i.e. integers) here should still have an edge by a couple of factors.


utils.py
```diff
@@ -0,0 +1,551 @@
+"""Lifetimes utils and helpers."""
```
CamDavidsonPilon: hm, did you mean to edit the lifetimes.utils python file? As is, this PR adds a brand new `utils.py` file.

MichaelSchreier: That's what you get for being lazy and using the web interface rather than a proper client -.-
This is fixed, now though there seem to be some some residual issues with period values which I will fix asap.

MichaelSchreier:
For compatibility reasons `_find_first_transactions` has to return `datetime_col` as `pd.Period`. This requires castings the timestamps back to periods before returning the results. This cast turns out to be relatively expensive, degrading the overall performance by a factor of 2-3 (on very large datasets) compared to when no cast is made.
The implementation remains substantially faster than before and probably faster still after pandas 0.24.0 will have addressed some of the performance issues with period values.

CamDavidsonPilon:
Thanks for the performance additions @MichaelSchreier! lgtm!

PR #243, opened 2019-01-23
add support for monthly and weekly sampled data in clv calculation

### LINKED ISSUE #233: _customer_lifetime_value implementation of DCF does not consider resampling freq of the transactions
```python
def _customer_lifetime_value(transaction_prediction_model, frequency, recency, T, monetary_value, time=12, discount_rate=0.01):

    df = pd.DataFrame(index=frequency.index)
    df['clv'] = 0  # initialize the clv column to zeros

    for i in range(30, (time * 30) + 1, 30):
        # since the prediction of number of transactions is cumulative, we have to subtract off the previous periods
        expected_number_of_transactions = transaction_prediction_model.predict(i, frequency, recency, T) - transaction_prediction_model.predict(i - 30, frequency, recency, T)
        # sum up the CLV estimates of all of the periods
        df['clv'] += (monetary_value * expected_number_of_transactions) / (1 + discount_rate) ** (i / 30)

    return df['clv']  # return as a series
```
In the the function assums that the frequency of purchases are always sampled per day i.e. freq="D". What if you have generated the summary using freq="M"? This will lead to wrong calculation inside the loop of _customer_lifetime_value().

ashish-ram:
fixes #233

### DIFF
--- a/lifetimes/fitters/gamma_gamma_fitter.py
+++ b/lifetimes/fitters/gamma_gamma_fitter.py
@@ -52,15 +52,14 @@ class GammaGammaFitter(BaseFitter):
     @staticmethod
     def _negative_log_likelihood(params, frequency, avg_monetary_value,
                                  penalizer_coef=0, q_constraint=False):
-            
-       
+
         if any(i < 0 for i in params):
             return np.inf
 
         p, q, v = params
         if q_constraint and q < 1:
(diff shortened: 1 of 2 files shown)


CamDavidsonPilon:
Interesting additions - are you able to add tests too?

PR #248, opened 2019-02-08
fix bug with hardcode date parameter

### LINKED ISSUE #217: expected_cumulative_transactions hardcodes datetime_col as 'date'
I think there's a bug in the function `expected_cumulative_transactions` that prevents causes an error if the datetime_col in the transactions dataframe is not 'date'. It also comes up if you try to run the `plot_cumulative_transactions` function. You can generate the error by running the following:

```
from lifetimes import ModifiedBetaGeoFitter
import lifetimes.plotting as ltplt
from lifetimes.utils import summary_data_from_transaction_data, expected_cumulative_transactions

transactions = load_transaction_data().rename(columns={'date':'Date'})
summary = summary_data_from_transaction_data(transactions,'id','Date')
mbg = ModifiedBetaGeoFitter()
mbg.fit(summary['frequency'],summary['recency'],summary['T'])
expected_cumulative_transactions(mbg, transactions, 'Date', 'id', 100)
```

This returns the error "Key Error: 'date'", which is triggered by line 487 in utils.py:
```
first_trans_size = first_transactions.groupby('date').size()
```

It looks to me as though 'date' is hardcoded on lines 487 and 499. I think in both cases it should use datetime_col instead, although I am not certain about line 499.

The function works fine if you run:
```
expected_cumulative_transactions(mbg,
                                 transactions.rename(columns={'Date':'date'}),
                                 'date', 'id', 100)
```

aprotopopov:
Fix silly bug for `'date'`. Closes #217

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -495,7 +495,7 @@ def expected_cumulative_transactions(model, transactions, datetime_col,
     date_periods = date_range.to_period(freq)
 
     pred_cum_transactions = []
-    first_trans_size = first_transactions.groupby('date').size()
+    first_trans_size = first_transactions.groupby(datetime_col).size()
     for i, period in enumerate(date_periods):
         if i % freq_multiplier == 0 and i > 0:
             times = period - first_trans_size.index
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
lgtm, circle is failing due to pandas 0.24 incompatibility I believe

Thanks @aprotopopov!

Issue #263, opened 2019-04-15
plot_calibration_purchases_vs_holdout_purchases taking for ever to run


amrishan:
![image](https://user-images.githubusercontent.com/36165834/56150244-12564700-5f74-11e9-84f2-fd36db3c37e8.png)

amrishan:
I don't think the lambda function is necessary. It's looping each value to make a prediction.

CamDavidsonPilon:
Riiight, this looks like a big perf win. Are you comfortable submitting a PR, or would you like to leave it to me?

amrishan:
Sure, I would be happy to assist with the PR. Thanks for the awesome package @CamDavidsonPilon .

amrishan:
> Riiight, this looks like a big perf win. Are you comfortable submitting a PR, or would you like to leave it to me?

Sure, I would be happy to assist with the PR. Thanks for the awesome package @CamDavidsonPilon .

PR #266, opened 2019-04-22
speedup/plot_calibration_purchases_vs_holdout_purchases

amrishan:
This is an implementation to the enhancement filed over at https://github.com/CamDavidsonPilon/lifetimes/issues/263 to replace lambda function with a direct call to model.conditional_expected_number_of_purchases_up_to_time method for performance improvement.


CamDavidsonPilon:
awesome, thanks!

Issue #274, opened 2019-05-30
scipy==1.2.0 causes dependency errors


davidpbowe:
Installing lifetimes with python3.7 caused the below dependency error.

`Traceback (most recent call last):
  File "UserStatus.py", line 9, in <module>
    from lifetimes.datasets import load_transaction_data
  File "~/.virtualenvs/analysis/lib/python3.7/site-packages/lifetimes/__init__.py", line 5, in <module>
    from .fitters.beta_geo_fitter import BetaGeoFitter
  File "~/.virtualenvs/analysis/lib/python3.7/site-packages/lifetimes/fitters/beta_geo_fitter.py", line 9, in <module>
    from autograd.scipy.special import gammaln, beta, gamma
  File "~/.virtualenvs/analysis/lib/python3.7/site-packages/autograd/scipy/__init__.py", line 7, in <module>
    from . import misc
  File "~/.virtualenvs/analysis/lib/python3.7/site-packages/autograd/scipy/misc.py", line 7, in <module>
    logsumexp = primitive(scipy.misc.logsumexp)
AttributeError: module 'scipy.misc' has no attribute 'logsumexp'`

Pip install seemed to have installed scipy==1.2.0 where the location of the logsumexp function has changed. Downgrading to scipy==1.1.0 solves the issue

teresadong:
I believe scipy.misc.logsumexp is deprecated according to: https://github.com/HIPS/autograd/issues/501
scipy.special.logsumexp I think is the correct new path

aprotopopov:
Due to that error, I suggest modifying requirements.txt for now for the autograd package to install it from git from [fix commit](https://github.com/HIPS/autograd/commit/4ec5aede380c5accf20afd4f4a96d59f1a9cd1df). 

```
-e git://github.com/HIPS/autograd.git@4ec5aede380c5accf20afd4f4a96d59f1a9cd1df#egg=autograd
```

PR #279, opened 2019-06-19
Fix autograd error by installing from git

aprotopopov:
Temporary fix for #274.


CamDavidsonPilon:
I'm okay with this (and I've been watching autograd's dev closely to track a release). Another solution is to pin scipy to <1.3.

Let's try this solution though

Issue #281, opened 2019-06-20
Bug in `calibration_and_holdout_data`


NudnikShpilkis:
There's a bug in `calibration_and_holdout_data` related to the bug in #280 .  `delta_time` is not properly calculated to account for periods.  In the example below, I try to find the `delta_time` between `2019-06-10` and`2019-04-30`, with a period of `7D`.  It should be ~5.87 since there 41 days in between those dates and 41/7 ~  5.87.  Instead, it returns 41.

```python
from  datetime import datetime
import numpy as np

(pd.to_datetime("2019-06-10").to_period("7D") - pd.to_datetime("2019-04-30").to_period("7D")).n
---------------------------------------------------------------------------------------------------------
41

(pd.to_datetime("2019-06-10").to_period("7D") - pd.to_datetime("2019-04-30").to_period("7D")) / np.timedelta64(1, '7D')
---------------------------------------------------------------------------------------------------------
5.857142857142857
```

Here is a fixed version of the function.  We can also remove `to_period` and assign the period to the column directly.  This greatly speeds up calculations in my experiments.
```python
def calibration_and_holdout_data2(
    transactions,
    customer_id_col,
    datetime_col,
    calibration_period_end,
    observation_period_end=None,
    freq="D",
    datetime_format=None,
    monetary_value_col=None,
):
    if observation_period_end is None:
        observation_period_end = transactions[datetime_col].max()

    transaction_cols = [customer_id_col, datetime_col]
    if monetary_value_col:
        transaction_cols.append(monetary_value_col)
    transactions = transactions[transaction_cols].copy()

    transactions[datetime_col] = pd.to_datetime(transactions[datetime_col], format=datetime_format)
    observation_period_end = pd.to_datetime(observation_period_end, format=datetime_format)
    calibration_period_end = pd.to_datetime(calibration_period_end, format=datetime_format)

    # create calibration dataset
    calibration_transactions = transactions.loc[transactions[datetime_col] <= calibration_period_end]
    calibration_summary_data = summary_data_from_transaction_data(
        calibration_transactions,
        customer_id_col,
        datetime_col,
        datetime_format=datetime_format,
        observation_period_end=calibration_period_end,
        freq=freq,
        monetary_value_col=monetary_value_col,
    )
    calibration_summary_data.columns = [c + "_cal" for c in calibration_summary_data.columns]

    # create holdout dataset
    holdout_transactions = transactions.loc[
[...]

Issue #280, opened 2019-06-19
plot_cumulative_transactions behaving as expected?


NudnikShpilkis:
I think there are a few bugs in the calculations for both the predicated and actual cumulative transactions in `expected_cumulative_transactions` that propagate to `plot_cumulative_transactions`.  I've recreated an example below using [cdnow_sample](http://www.brucehardie.com/datasets/CDNOW_sample.zip).

```python
from  datetime import datetime

import numpy as np
import pandas as pd

from lifetimes import ModifiedBetaGeoFitter
from lifetimes.utils import calibration_and_holdout_data, expected_cumulative_transactions
from lifetimes.plotting import *

import matplotlib.pyplot as plt

df= pd.read_csv("cdnow_sample.csv")
df['Date'] = pd.to_datetime(df["Date"])

df_rfm = calibration_and_holdout_data(df, 
                            customer_id_col="ID",
                            datetime_col="Date",
                            calibration_period_end=datetime(1997, 9, 30),
                            freq='7D',
                            monetary_value_col='Spend')

df_rfm.head()

mbg = ModifiedBetaGeoFitter(penalizer_coef=0.0).fit(df_rfm['frequency_cal'], df_rfm['recency_cal'], df_rfm['T_cal'])
mbg.summary
-----------------------------------------------------------------------------------------------------

coef | se(coef) | lower 95% bound | upper 95% bound
-- | -- | -- | --
0.524843 | 0.083268 | 0.361637 | 0.688048
6.183082 | 0.695620 | 4.819667 | 7.546497
0.891360 | 0.154790 | 0.587971 | 1.194748
1.614003 | 0.537495 | 0.560513 | 2.667493

plot_period_transactions(mbg, max_frequency=10)
```
![image](https://user-images.githubusercontent.com/17383024/59803664-2b53df00-92ba-11e9-97dd-930cbc6d70e3.png)

The plot below is generated using `plot_cumulative_transactions`.  The predicted transaction total is larger than the cumulative actual dataset after 1 period.
```python
plot_cumulative_transactions(mbg, 
                             transactions=data,
                             datetime_col="Date",
                             customer_id_col="ID",
                             t=10,
                             t_cal=5,
                             freq='7D')
```
![image](https://user-images.githubusercontent.com/17383024/59800157-dc0ab000-92b3-11e9-8ffc-167eae7d3d83.png)

[...]

Issue #259, opened 2019-03-21
BetaGeoBetaBinomFitter.fit() not converging


cereusperuv:
Any ideas what might be going on? Or more likely, what I've done wrong... Main parts of the error below. Thanks!

Raised by `BaseFitter._fit()`:

`ConvergenceError: The model did not converge. Try adding a larger penalizer to see if that helps convergence.`

`print(output)` in `BaseFitter._fit()`:

```
      fun: nan
 hess_inv: array([[1, 0, 0, 0],
       [0, 1, 0, 0],
       [0, 0, 1, 0],
       [0, 0, 0, 1]])
      jac: array([nan, nan, nan, nan])
  message: 'Desired error not necessarily achieved due to precision loss.'
     nfev: 1
      nit: 0
     njev: 1
   status: 2
  success: False
        x: array([0.1, 0.1, 0.1, 0.1])
```

Autograd warnings:

```
C:\Tools\Anaconda3\envs\TestEnv\lib\site-packages\autograd\tracer.py:48: RuntimeWarning: invalid value encountered in log
  return f_raw(*args, **kwargs)
C:\Tools\Anaconda3\envs\TestEnv\lib\site-packages\autograd\tracer.py:48: RuntimeWarning: invalid value encountered in logaddexp
  return f_raw(*args, **kwargs)
```

Numpy warning:

```
C:\Tools\Anaconda3\envs\TestEnv\lib\site-packages\numpy\core\fromnumeric.py:83: RuntimeWarning: invalid value encountered in reduce
  return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
```

CamDavidsonPilon:
Hey @cereusperuv, thanks for the report. Is is possible for you to send the dataset over (email is fine for me), or can you answer the following:

1. what is the length of the dataset?
2. What is the maximum value of frequency?
3. Also, does adding a small penalizer help? BetaGeoBetaBinomFitter(penalizer=0.001)?

cereusperuv:
Hi @CamDavidsonPilon and thanks for your prompt reply. I will send you my dataset shortly, just have to edit it a tad first. Meanwhile, I can tell you that

1. The pre-processed dataset on form (id, frequency, recency, T) has length 24308.
2. The maximum frequency is 152.
3. A small penalizer does not alter the situation. I have tried the values 0.001, 0.01, 0.1, 1.

CamDavidsonPilon:
Thanks.

[...]

PR #287, opened 2019-07-01
Fixed periodicity in rfm utils

NudnikShpilkis:
I've written some changes to the following functions to better align them with the suspected bugs per #280 and #281:

- `calibration_and_holdout_data`: I've added variables to count the first-period transactions, calculate recency for discrete time models (as per https://github.com/CamDavidsonPilon/lifetimes/issues/280#issuecomment-505526696).  I'm not convinced that counting the first-period transactions follows the literature, however, we can let users decide what they'd prefer.  I've added an additional variable to let the user calculate recency as a discrete measure for use in the BG/BB (as per http://www.brucehardie.com/notes/010/).  I've also separated out the operations to generate holdout data into its own function for better testing and model validation.

- `summary_data_from_transaction_data`: Grouping is now done by `customer_id_col` and a special Pandas grouper based on the user-defined frequency.  The function takes the same additional inputs that `calibration_and_holdout_data` takes.

- `expected_cumulative_transactions`: The dates and summed aggregated transactions are now calculated correctly using the `freq` and start date.

`beta_geo_beta_binomial`: Updated `fit` function as per https://github.com/CamDavidsonPilon/lifetimes/issues/259#issuecomment-480820040.  Additionally aliased `predict` function as `conditional_expected_number_of_purchases_up_to_time` to make the fitter compatible with a few more plotting functions.

I've added the following functions:

- `holdout_data`: I've split off the holdout data calculations from `calibration_and_holdout_data` into their own function.  I found it easier to have a smaller function to calculate the holdout data for model validation.
- `plot_calibration_purchases_vs_holdout_spend`: Almost identical to `plot_calibration_purchases_vs_holdout_purchases`.  The Y-axis plots the CLV instead of purchases in the holdout period.  An example of this plot can be seen in http://www.brucehardie.com/papers/020/fader_et_al_mksc_10.pdf.

I have not yet written the unit tests for these changes, but I wanted to give others the opportunity to view these changes and comment.


CamDavidsonPilon:
Thoughtful additions, thanks! 

There are a number of files under `lifetimes/.ipynb_checkpoints` that shouldn't be there. Can those be removed from the PR please?

isaachaberman:
Will do!  

I'm also working through the tests now to make sure all the changes are up to snuff.

isaachaberman:
@CamDavidsonPilon How should we treat `observation_period_end` in the utils functions.  If we treat it as the specific day itself regardless of the period we'll be more exacting.  However, our calculations for T will not be integers.  If we treat it as a period, our results will be less exacting and T will always be an integer. 

Here's a short example exploring the differences:

```python
import numpy as np
import pandas as pd

observation_period_end = "2015-02-07"
calibration_end = "2015-02-01"
freq, freq_multiplier = "W", 1

#With periods
(pd.to_datetime(today).to_period('W') - pd.to_datetime(calibration_end).to_period('W')).n
1

#With exact dates
(pd.to_datetime(today) - pd.to_datetime(calibration_end)) / np.timedelta64(1, freq) / freq_multiplier
0.8571428571428571

```

CamDavidsonPilon:
No reason for T to be an integer (do you agree?). I'm happy with the former.

isaachaberman:
I prefer the former as well as it keeps to a stricter definition of periods.

isaachaberman:
I've added a few more functions to plot the cumulative and incremental CLV.  To do so, I added some new utils functions including a `customer_lifetime_value` function that should work with any `freq` not just months.  Let me know what you think.

isaachaberman:
I've added one new function `plot_cumulative_clv_from_t_cal` which lets you plot cumulative clv from the end of the calibration period.  It uses `expected_cumulative_clv` and `customer_lifetime_value` as the other cumulative CLV functions do, but requires the user to enter the calibration period ending and a starting CLV value for all customers.

psygo:
So... again, I haven't had much time lately, but I'm going to do my best to schedule one or two days next week to incorporate the PR queue into the `dev` branch at least.

But, in order to ease our lives &mdash; maybe one day you will review my PRs also &mdash;, I would ask you (@isaachaberman) to do some things for me:

1. Getting the `.ipynb_checkpoints/` folders out of the way by adding them to the `.gitignore` file, as they are mostly Jupyter logs.
    - You can get them in every possible folders all at once by doing something like `**/.ipynb_checkpoints` I think.
1. I have been thinking about adding a `benchmarks` folder to the library where we would store scripts that demonstrate a bug or show how to use some special function, etc.
    - You could then put one for the bug(s) you found in there.
1. Create a separate PR for the fixes or modifications of the `plotting` functions. This would ease the life of whoever is trying to review the code.

If you're enthusiastic about this library, I would for sure welcome you to the team as a new collaborator &mdash; I'm a new member as well. You would only need the @CamDavidsonPilon's permission. The next thing/issue I really think (I) should be taken care of are the *convergence errors* and how to deal with the *negative log special cases*, if you can take a look at them, it would be very valuable.

PR #289, opened 2019-07-02
Added a file (docs) for a Higher Level Overview of the Library

psygo:
This is supposed to be a first sketch at a more friendly documentation. Hopefully, it will also help easing future contributions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ target/
 
 # PyCharm
 .idea/*
+
+# VS Code
+.vscode/
--- a/docs/High Level Overview.md
+++ b/docs/High Level Overview.md
@@ -1,0 +1,48 @@
+# High Level Overview
+
+This is intended to be a high-level documentation of how the code is structured. Whenever possible, [UML](https://en.wikipedia.org/wiki/Unified_Modeling_Language) is used. Some of the standards applied in this documentation can be found [here](https://www.lucidchart.com/pages/uml-class-diagram).
+
+## Workflow
+
+The usual workflow of using the `Lifetimes` library is exemplified in the [Quickstart](Quickstart.md) page. It can also be represented through the following fluxogram:
+
+![Basic Workflow](https://i.imgur.com/oV2KpQG.png)
+
+Notice that the right-most branch of the fluxogram actually refers to *monetary value* modeling.
+
+## Fitters
+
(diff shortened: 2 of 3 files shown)


psygo:
(Sorry, but I accidentally made this a pull request when it should have been better to make it a *draft pull request*. This is also my first contribution on Github ever, so, please, if you have pro suggestions, don't hesitate to give them.)

So... I had (still have) some difficulties trying to understand how things work with this library. I think it's not an easy area initially, especially when trying to grasp how to organize everything in your head &mdash; even considering the extremely helpful `Quickstart` docs page. That's why I created most of what is on the `High Level Overview` page as a way of studying; and later decided to transform it into a pull request.

Since documentation is something very tied to the author's style of writing, I expect this to be a first suggestion of what the page could look like. I tried to follow some of the conventions implicitly defined in the `Quickstart` page, but who knows.

Most of the fluxograms were created with [Lucidchart](https://www.lucidchart.com/), the shareable link can be hopefully found [here](https://www.lucidchart.com/invitations/accept/67fedd68-afbd-4ba5-96d6-8e62f3c180d0).

psygo:
Two items I would like to include in the future that I don't think are very intuitive to those first studying the related topics are:
- I believe the two main articles of the library should be prominently featured on the main `README.md` (and the docs web page). Schmittlein's article deserves special mention because it's paid and, thus, people will initially not be inclined to read it &mdash; besides that, it makes reading Fader's paper much easier. The two articles are:
    - *"Counting Your Customers" the Easy Way: An Alternative to the Pareto/NBD Model*, by P. Fader, B. Hardie and K. Lee.
    - *Counting Your Customers: Who Are They and What Will They Do Next*, by D. Schmittlein, D. Morrison and R. Colombo.
- This is something that still is a major point of confusion to me and certainly will be to anyone who enters this field: the probability of a certain customer having `x` purchases is *actually* the "crude" probability of him having `x` purchases *times* the probability of him being alive at the specific point in time. So, the question is:
    - How was that implemented in the library? Is it through prepending the purchase probabilities with the term *condtional*?

CamDavidsonPilon:
This is great, @psygo! Thanks for these additions (and great idea)

CamDavidsonPilon:
Also, congrats on your first PR! 🎉

psygo:
Cool, nice that it is in now and that you liked it.

But what about the two other suggestions/questions I mentioned? And were you able to open the Lucidchart fluxograms I mentioned? Sorry if you're too busy :P.

PR #290, opened 2019-07-02
Better Docstrings for the (some) Fitters

psygo:
The modified files are the *BG/NBD*, *Pareto/NBD* and *Gamma-Gamma* fitters. I've also added some stuff to the `utils.py` file.

The intention behind this pull request is essentially to add mentions to the equations behind some methods and functions. However, I've also reformatted some of the code, nothing beyond superficial tabs and line brakes and spacings to add a bit more consistency to the files.

If this is formatting is welcome, I could do it for the rest of the code base; else, I can revert what I changed and submit only the docstring improvements.

I intend on fixing some minor docstring (and hard coding) issues I've also seen in the `plotting.py` file next, so please let me know if I should keep going or simply stop.

### DIFF
--- a/lifetimes/fitters/beta_geo_fitter.py
+++ b/lifetimes/fitters/beta_geo_fitter.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Beta Geo Fitter, also known as BG/NBD model."""
+
 from __future__ import print_function
 from __future__ import division
 import warnings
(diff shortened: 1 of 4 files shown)


psygo:
Btw, I still haven't understood why is it that you're importing `matplotlib.pyplot` in every function instead of at the top of the file. Is it because you didn't want to import it when the file is imported because it's a big library?

psygo:
Take a look at the [related plot *PR*](https://github.com/CamDavidsonPilon/lifetimes/pull/291).

CamDavidsonPilon:
> Btw, I still haven't understood why is it that you're importing matplotlib.pyplot in every function instead of at the top of the file. Is it because you didn't want to import it when the file is imported because it's a big library?

I think it's to make matplotlib an optional dependency. I think, however, it's popular enough that we can probably promote it to a required dependency.

CamDavidsonPilon:
> I intend on fixing some minor docstring (and hard coding) issues I've also seen in the plotting.py file next, so please let me know if I should keep going or simply stop.

Please continue, these are helpful additions

PR #295, opened 2019-07-08
Log Likelihood Trackers for Convergence

psygo:
Essentially, this *PR* creates **fitter properties that track the parameters' convergence**. This way the user will be able to debug the fitting visually and numerically.

Other subfeatures introduced are:

1. A new `benchmarks` folder with scripts that would test the benchmarked use cases, such as the one on the `Quickstart` web page and the `CDNOW` dataset. On later *PR*s I will store useful scripts there.
    - The `images` subfolder is ignored within the `.gitignore` file.
1. Plots for the fitters' parameters before and after exponentiation were added.

On a sidenote, I would like to point out the somewhat weird behavior I've obtained for the Pareto/NBD fitter on the `Quickstart` artificial data &mdash; maybe that's why Fader complains about fitting it? &mdash;:

Pareto/NBD Parameters Convergence

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ target/
 
 # VS Code
 .vscode/
+
+# Benchmarks Images
+benchmarks/images/
--- a/benchmarks/convergence_summary_graphs.py
+++ b/benchmarks/convergence_summary_graphs.py
@@ -5,7 +5,10 @@
 import sys
 
 import lifetimes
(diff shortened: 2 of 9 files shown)


psygo:
Inside the `benchmarks`, the `convergence_summary_graphs.py` script doesn't have full `Python 2.x` support, basically due to the `print`s though. 

I've used `range` inside one of the plots, do I need to replace it because of `Python 2.x`?

PR #329, opened 2019-10-25
Clarify that monetary_value is defined as a mean instead of a sum in the docs

airhorns:
This tripped me up for a while, figured it'd be nice to help out the next person that comes along and makes the same bad assumption :) 

For anyone else that discovers this later, this is often a cause of the correlation between `monetary_value` and `frequency` being much higher than you might expect!


CamDavidsonPilon:
Nice addition 👍 

Thanks @airhorns!

clausherther:
Thanks @airhorns !

PR #375, opened 2020-06-10
utils.py/calibration_and_holdout_data: Add freq_multiplier

utkarshgupta137:
Added freq_multiplier to calibration_and_holdout_data.
Also updated 'freq' definition with the correct link for possible values and 'freq_multiplier' definition to make it clearer.

### DIFF
--- a/lifetimes/utils.py
+++ b/lifetimes/utils.py
@@ -31,6 +31,7 @@ def calibration_and_holdout_data(
     calibration_period_end,
     observation_period_end=None,
     freq="D",
+    freq_multiplier=1,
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
I'm confused about why freq multiplier is needed?

utkarshgupta137:
This function is useful for splitting data, after which it can be used with other libraries as well. Since this option is already provided with summary_from_data, I see no harm in providing it in calibration_and_holdout_data

CamDavidsonPilon:
Makes sense 👍 thanks @utkarshgupta137!

Issue #262, opened 2019-04-12
convergence errors with v0.11.1


sophia-wright-blue:
hello @CamDavidsonPilon , i was able to use lifetimes v0.10.1 perfectly on my data (customer purchase data for retail stores), but have been getting convergence errors on v0.11.1. I see that others have been getting convergence errors too, and in response you ask questions about:

1). length of the dataset

2). max value of frequency

3). does adding a small penalizer help? BetaGeoFitter(penalizer=0.001)?

my question is: what about these 3 attributes is causing the convergence errors in the new version and not the old version? are there any limits to the values for 1) and 2) above? I'm trying to figure out if there is any additional data prep that I can do to avoid the convergence errors;

I'd greatly appreciate a response - can't thank you enough for the amazing work that you have done to make this library available.

=========

psygo:
I don't know exactly what's breaking for you in this specific version transition &mdash; maybe you could post part of your data set so others can test?

The penalizer should help convergence most of the time. But it also might help it not converge if it is too big. I usually use it from `0` to `0.5`; the graphs with the calibration vs holdout purchases tell me if I'm too far off the target.

sophia-wright-blue:
thanks for replying @psygo , I've fallen behind on the projects I had planned with the `lifetimes` library - I'll try and post some data so others can recreate the same error that I'm getting. However, quite a few people have raised an issue about getting convergence errors with `v0.11.1` and not with `v0.10.1`.

I could resolve the convergence error by using the `penalizer`, but I'm extremely interested in finding the underlying reason behind the convergence error, I'd greatly appreciate it if you could shed some light on the theory that'd explain this error

psygo:
That's something I'm going to check out as soon as possible, as it might be a hidden bug for me too.

In theory, it shouldn't be *that hard* to find out where the bug is, since it is probably a Git discernable difference in the `BaseFitter` class. But, of course, practica may prove otherwise.

sophia-wright-blue:
thanks for your reply @psygo , I'm going to track the progress in the `Correct Convergence Errors on Version 0.11.1` milestone and will try and find some time to understand the problem better

PR #381, opened 2020-07-06
More numerically stable log likelihood calculation for beta_geo_fitter

Nicky027:
I ran into issues when using BetaGeoFitter from v0.11.1. 

The parameter estimates I was getting did not make sense (I was getting r=a=b=1.105171 and alpha=2295.439997 and NaN for se(coef) when looking at the summary). 

The reason for this was that in the computation of `ll` in the `_negative_log_likelihood` it was hitting inf for some values when computing `np.exp(A_3)` and `np.exp(A_4)`. Since we eventually compute the log of the sum of these two terms, we can avoid hitting inf by subtracting the largest of A_3 and A_4 when doing np.exp() and then adding it back outside of the np.log().  This leaves the expression mathematically the same, but since np.exp() is done on much smaller numbers, then you would avoid running into the numerical issues caused by doing np.exp() on large numbers.

This change is possibly a solution for this currently open issue: https://github.com/CamDavidsonPilon/lifetimes/issues/262

### DIFF
--- a/lifetimes/fitters/beta_geo_fitter.py
+++ b/lifetimes/fitters/beta_geo_fitter.py
@@ -191,8 +191,10 @@ class BetaGeoFitter(BaseFitter):
         A_3 = -(r + freq) * np.log(alpha + T)
         A_4 = np.log(a) - np.log(b + np.maximum(freq, 1) - 1) - (r + freq) * np.log(rec + alpha)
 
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
This is so obvious - I love it. Have you seen an improvement with these changes?

Nicky027:
Indeed. I got proper estimates after these changes 🙂

PR #385, opened 2020-07-06
Fix infavorable numpy version in build env

Arnoldosmium:
Force updating the cached pip dep environment in travis seems to fix the py3.6 build.
The former cached `numpy` version was too old (1.13.3) compared to other envs in the build matrix (1.15.0)

We can add more version pinning if needed, like the pin for `pandas`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
    - "3.5"
    - "3.6"
 env:
-   - PANDAS_VERSION=0.24.1
+  - PANDAS_VERSION=0.24.1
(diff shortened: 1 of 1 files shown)


CamDavidsonPilon:
Great, thanks! I was having a similar problem with *lifelines* and I think this work there, too

CamDavidsonPilon:
Generally, I need to update/check the deps on this library, too

Arnoldosmium:
Indeed, updating the minimum version requirement of dependencies can avoid some build / user install issues. It could shrink the supported numpy / pandas version range though.

PR #399, opened 2020-10-04
update requirements.txt and dev_requirement.txt

rahulshivan05:


CamDavidsonPilon:
Hi @rahulshivan05,

why do you think this is a good idea?

